PR #11, opened 2022-10-14
Update outcome status label and move `Status` enum

janstenpickle:
Changes the label name to reduce the possibility of a clash with other labels.

Moves `Status` class under `OutcomeRecorder` as it is only used there.


TimWSpence (approved):
`outcome_status` does feel like saying the same thing twice but I can't think of a better label name. Unless `status` but that's very vague

janstenpickle:
> `outcome_status` does feel like saying the same thing twice but I can't think of a better label name. Unless `status` but that's very vague

There's time before the proper release, so we can review!

bastewart:
Ah, to make it a bit less tautology-y potentially `recorded_status`?

PR #15, opened 2022-10-19
Add callback functionality

janstenpickle:


TimWSpence (approved):
I still feel like there must be a better way to model this but I have no idea what it is 😭 Regardless, it's really nice that it works so good work!

PR #16, opened 2022-10-19
Add info metric type

janstenpickle:
This is a very special type, it just takes a map of labels which are then represented like gauges with just a value of `1`. In the Java library it _is_ possible to create a labelled version of this, where the label names and values can be provided separately, but this seems entirely pointless to me, as does having a callback version.

### DIFF
--- a/core/src/main/scala-2/prometheus4cats/InfoNameFromStringLiteral.scala
+++ b/core/src/main/scala-2/prometheus4cats/InfoNameFromStringLiteral.scala
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2022 Permutive
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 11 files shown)


core/src/main/scala-2/prometheus4cats/InfoNameFromStringLiteral.scala
```diff
@@ -0,0 +1,44 @@
+/*
```
TimWSpence, line 1: I assume this will be moved to `internal` like the others?

TimWSpence (approved):
LGTM!

core/src/main/scala-2/prometheus4cats/InfoNameFromStringLiteral.scala
janstenpickle: Yep, that's in the other PR

PR #20, opened 2022-10-21
Refinement cleanup

### LINKED ISSUE #18: TODO: Metric specific labels
Some metric types have rules about label names that must not be certain strings; e.g. `le` for histograms and `quantile` for summary.

janstenpickle:
Remove some of the boilerplate in refined types

Fixes #18

### DIFF
--- a/core/src/main/scala/prometheus4cats/Label.scala
+++ b/core/src/main/scala/prometheus4cats/Label.scala
@@ -16,64 +16,27 @@
 
 package prometheus4cats
 
-import cats.{Eq, Hash, Order, Show}
+import java.util.regex.Pattern
+
 import prometheus4cats.internal.LabelNameFromStringLiteral
 
 object Label {
 
   /** Refined value class for a label name that has been parsed from a string
     */
-  final class Name private (val value: String) extends AnyVal {
+  final class Name private (val value: String) extends AnyVal with internal.Refined.Value[String] {
 
     override def toString: String = s"""Label.Name("$value")"""
 
   }
 
-  object Name extends LabelNameFromStringLiteral {
-
-    final private val regex = "^[a-zA-Z_:][a-zA-Z0-9_:]*$".r.pattern
-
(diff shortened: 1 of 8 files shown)


core/src/main/scala/prometheus4cats/Label.scala
```diff
@@ -16,64 +16,35 @@
...
-
+  object Name extends internal.Refined[String, Name] with LabelNameFromStringLiteral {
     // prevents macro compilation problems with the status label
     private[prometheus4cats] val outcomeStatus = new Name("outcome_status")
 
-    implicit val catsInstances: Hash[Name] with Order[Name] with Show[Name] = new Hash[Name]
-      with Order[Name]
-      with Show[Name] {
-      override def hash(x: Name): Int = Hash[String].hash(x.value)
+    private val invalidNames: Set[String] = Set("quantile", "le", "name")
```
janstenpickle, line 37: Histograms require that no user defined label is `le` and summaries require the same for `quantile`. This raises the question: should we have metric specifc label names, just as we do for metric names?

bastewart: Yeah, I think if we can get this cleanly into the DSL (like it is for names) then definitely 👍

If we _can't_ then what I imagine you're doing now (runtime exception) is fine.

janstenpickle: This isn't actually a runtime execption... this is more a question about whether, like with metric names, we should have metric specific labels.

Speaking with @TimWSpence we think it's probably a good idea to go with metric specific for consitency's sake.

janstenpickle: I've just had a go at wiring this through and it is a bit of a pain. Probably not worth the effort considering there are only two labels that we want to prevent peoplefrom using...

TimWSpence (approved):
Nice!

PR #24, opened 2022-10-31
Metrics as resources

janstenpickle:
This returns each metric as a resource, meaning that it gets de-registered from the registry upon resource finalisation.

### DIFF
--- a/java/src/main/scala/prometheus4cats/javasimpleclient/internal/MetricCollectionProcessor.scala
+++ b/java/src/main/scala/prometheus4cats/javasimpleclient/internal/MetricCollectionProcessor.scala
@@ -18,18 +18,16 @@ package prometheus4cats.javasimpleclient.internal
 
 import java.util
 
+import alleycats.std.iterable._
 import alleycats.std.set._
-import cats.data.NonEmptyList
 import cats.effect.kernel._
 import cats.effect.std.Dispatcher
-import cats.syntax.applicativeError._
 import cats.syntax.flatMap._
 import cats.syntax.foldable._
 import cats.syntax.functor._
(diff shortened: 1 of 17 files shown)


java/src/main/scala/prometheus4cats/javasimpleclient/internal/MetricCollectionProcessor.scala
```diff
@@ -338,16 +354,11 @@ object MetricCollectionProcessor {
...
-      Logger[F].warn(e)(s"Failed to unregister a collector on shutdown.")
-    }
-
     Resource.make(acquire) { proc =>
-      handleError(Sync[F].delay(promRegistry.unregister(callbackHist))) >> handleError(
-        Sync[F].delay(
-          promRegistry.unregister(duplicateGauge)
-        )
-      ) >> handleError(Sync[F].delay(promRegistry.unregister(proc)))
+      Utils.unregister(callbackHist, promRegistry) >> Utils.unregister(
```
TimWSpence, line 358: Can these things throw? If so we need `guarantee` rather than `>>`. Same question for all the other occurrences, although most of them are just updating `Ref`s, which _should_ be safe

TimWSpence (approved):
Nice!

java/src/main/scala/prometheus4cats/javasimpleclient/internal/MetricCollectionProcessor.scala
janstenpickle: They can, the `unregister` uses `handleErrorWith` and logs the error. It doesn't help with cancellation, but I don't think that's worth worrying about on shutdown

PR #27, opened 2022-11-02
Fix intellij implicit resolution

janstenpickle:
IJ seems to not be able to find `InitLast` implicits when they are type bounds on methods, adding it an implicit parameter works fine.

This
image

Becomes 

image


TimWSpence (approved):
LOL

PR #28, opened 2022-11-03
Allow multiple values to be passed on a labelled callback

janstenpickle:
It struck me, while porting the event sourcing SDK to Prometheus4Cats, that when recording a labelled callback you're almost always going to want to return multiple values, otherwise you may as well just set the labels you want in `CommonLabels`


janstenpickle:
question: is `NonEmptyList` the best datastructure to use here or should we make it a `NonEmptySeq`?

TimWSpence (approved):
LGTM! I think `NonEmptyList` is fine - it has well-defined performance characteristics

PR #41, opened 2022-12-06
Testing implementation

TimWSpence:
An in-memory implementation of `MetricsFactory` for testing.

### DIFF
--- a/testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
+++ b/testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
@@ -33,6 +33,7 @@ class TestingMetricRegistry[F[_]](
             Some(1 -> counter) -> F.pure(
               Resource.make(F.pure(counter))(_ => release)
             )
+          // TODO do we need to check if we have already registered with the same name but different labels or type?
           case Some((n, c)) => Some(n + 1 -> c) -> F.pure(Resource.make(F.pure(c))(_ => release))
         }.flatten
       })
--- a/testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
+++ b/testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
@@ -7,11 +7,43 @@ import cats.effect.kernel._
 import cats.effect.std.MapRef
 import cats.data.NonEmptySeq
 import scala.concurrent.duration.FiniteDuration
+import TestingMetricRegistry._
 
 class TestingMetricRegistry[F[_]](
-    private val counters: MapRef[F, (Counter.Name, Metric.CommonLabels), Option[(Int, Counter[F, Double])]]
+    private val store: MapRef[F, (String, List[String]), Option[(Int, MetricType, Metric[Double])]]
 )(implicit F: Concurrent[F])
     extends DoubleMetricRegistry[F] {
+
+  private def store[M <: Metric[Double]](
+      name: String,
+      labels: List[String],
+      tpe: MetricType,
+      create: Ref[F, Double] => M
+  ): Resource[F, M] =
+    Resource
+      .eval(F.ref(0.0).flatMap { ref =>
+        val release =
+          store(name -> labels).update {
+            case None => throw new RuntimeException("This should be unreachable - our reference counting has a bug")
+            case Some((n, t, c)) => if (n == 1) None else Some((n - 1, t, c))
+          }
+
+        store(name -> labels).modify {
+          case None =>
+            val m = create(ref)
+            Some((1, tpe, m)) -> F.pure(
+              Resource.make(F.pure(m))(_ => release)
+            )
+          case Some((n, t, c)) =>
+            Some((n + 1, t, c)) -> F.pure(
+              Resource.make(
+                // Cast safe by construction
+                F.pure(c.asInstanceOf[M])
+              )(_ => release)
+            )
+        }.flatten
+      })
+      .flatten
 
   override protected[prometheus4cats] def createAndRegisterDoubleCounter(
       prefix: Option[Metric.Prefix],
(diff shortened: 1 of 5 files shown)


testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
```diff
@@ -0,0 +1,283 @@
...
+    extends DoubleMetricRegistry[F] {
+
+  def counterHistory(name: Counter.Name, commonLabels: Metric.CommonLabels): F[Option[Chain[Double]]] =
+    metricHistory(name.value, commonLabels.value.values.toList, MetricType.Counter)
+
+  def counterHistory(
+      name: Counter.Name,
+      commonLabels: Metric.CommonLabels,
+      labelNames: IndexedSeq[Label.Name]
+  ): F[Option[Chain[Double]]] =
```
janstenpickle, line 41: This may need a `labelValues` argument or return label values with the result

testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
```diff
@@ -0,0 +1,283 @@
...
+import cats.data.Chain
+import cats.effect.kernel._
+import cats.effect.std.MapRef
+import cats.data.NonEmptySeq
+import prometheus4cats.util.DoubleMetricRegistry
+import scala.concurrent.duration.FiniteDuration
+import TestingMetricRegistry._
+
+class TestingMetricRegistry[F[_]](
+    private val store: MapRef[F, (String, List[String]), Option[(Int, MetricType, Metric[Double], F[Chain[Double]])]]
```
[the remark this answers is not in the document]
TimWSpence: Yep fair :+1

testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
```diff
@@ -0,0 +1,426 @@
...
+      labelValues: List[String],
+      tpe: MetricType
+  ): F[Option[Chain[Double]]] =
+    underlying(name -> labelNames).get.flatMap(_.flatTraverse {
+      case (_, t, _, r) if t == tpe =>
+        r(labelValues).get.map(_.some)
+      case _ => Option.empty[Chain[Double]].pure[F]
+    })
+
+  private def prefixedName(prefix: Option[Metric.Prefix], name: String): String = prefix.fold(name)(p => s"${p}_$name")
```
janstenpickle, line 390: There's a utility in `core` that will format the prefixed name for you

testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
```diff
@@ -0,0 +1,426 @@
...
+import cats.data.NonEmptySeq
+import prometheus4cats.util.DoubleMetricRegistry
+import scala.concurrent.duration.FiniteDuration
+import TestingMetricRegistry._
+
+class TestingMetricRegistry[F[_]](
+    val underlying: MapRef[F, (String, List[String]), Option[
+      (Int, MetricType, Metric[Double], MapRef[F, List[String], Chain[Double]])
+    ]],
+    val info: MapRef[F, String, Option[(Int, Info[F, Map[Label.Name, String]])]]
```
janstenpickle, line 33: No biggy, but `MapRef`s make me nervous because you can't get all of the contained values. That said I'm not sure how important doing that really is. What do you think?

testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
```diff
@@ -0,0 +1,426 @@
...
+                    F.pure(
+                      Resource.make(
+                        // Cast safe by construction
+                        F.pure(m.asInstanceOf[M])
+                      )(_ => release)
+                    )
+                else
+                  curr -> F.raiseError[Resource[F, M]](
+                    new RuntimeException(
+                      s"Cannot create metric of type $tpe as metric of type $t alreasy exists with the same name and labels"
```
janstenpickle, line 370: In the long term we want to support the ability to have a metric with the same name and different labels. I'm wondering if we do that here with the testing impl or wait until we have a better alternative.

janstenpickle: Also typo in `alreasy`

testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
```diff
@@ -0,0 +1,426 @@
...
+import cats.syntax.all._
+import cats.data.Chain
+import cats.effect.kernel._
+import cats.effect.std.MapRef
+import cats.data.NonEmptySeq
+import prometheus4cats.util.DoubleMetricRegistry
+import scala.concurrent.duration.FiniteDuration
+import TestingMetricRegistry._
+
+class TestingMetricRegistry[F[_]](
```
janstenpickle, line 29: I think this constructor could do with being private so that the data structure can change in the future if needed.

I realise this would mean it's harder to get at the underlying `MapRef`s, but I wonder if it's possible to expose the metric collections in a generic way that doesn't tie us to the implementation. I release that might conflict with my earlier review comment asking you to expose the underlying data, so if it's too much of a pain exposing the data, then I'd say not bother with lower level access and just provide the higher level methods.

TimWSpence: Yeah maybe let's start with just exposing our higher level methods and see if people come up with other usecases

testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
TimWSpence: Hmm good question. Related to the discussion below, it is now private so I guess we can keep it for the moment and then if people have a use-case for it then we can always replace it with `Ref[F, Map[...]]`?

testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
janstenpickle: Yep, that makes sense. It's much easier to add with bincompat, even if it's class that extends this. Which makes me wonder, should this be `sealed abstract`?

testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
TimWSpence: Ann I missing something? 😅 What is wrong with the message sorry?

testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
TimWSpence: Yeah makes sense 👍

testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
janstenpickle: Oh no, it's me 🤦. What you have is correct!

janstenpickle (approved):
LGTM!

PR #42, opened 2022-12-08
Move implicit evidence for DSL syntax

### LINKED ISSUE #40: Move imlpicit evidence for extra DSL to methods
See [`asTimer`](https://github.com/permutive-engineering/prometheus4cats/blob/main/core/src/main/scala/prometheus4cats/internal/package.scala#L49) for an example.

This means the methods would always show up and complain about missing implicits rather than not showing up at all if the implicit isn't in scope.

janstenpickle:
This ensures the syntax always shows up even if the user doesn't have the implicit in scope.

Fixes #40


TimWSpence (approved):
LGTM!

PR #54, opened 2023-01-11
Fix munit-cats-effect-3 being in the `compile` scope

bastewart:


bastewart:
Closing in favour of #53 which I had missed 🤦

PR #56, opened 2023-01-19
Remove check for duplicates in collection.

janstenpickle:
Also check that callbacks don't have duplicates

### DIFF
--- a/java/src/main/scala/prometheus4cats/javasimpleclient/JavaMetricRegistry.scala
+++ b/java/src/main/scala/prometheus4cats/javasimpleclient/JavaMetricRegistry.scala
@@ -771,7 +771,7 @@ object JavaMetricRegistry {
       metricCollectionProcessor <- MetricCollectionProcessor
         .create(
           ref,
-          // TODO callabck statee
+          callbackState,
           dis,
           metricCollectionCallbackTimeout,
           promRegistry
--- a/java/src/main/scala/prometheus4cats/javasimpleclient/JavaMetricRegistry.scala
+++ b/java/src/main/scala/prometheus4cats/javasimpleclient/JavaMetricRegistry.scala
@@ -20,23 +20,12 @@ import java.util
 
 import cats.data.{NonEmptyList, NonEmptySeq}
 import cats.effect.kernel._
+import cats.effect.kernel.syntax.monadCancel._
 import cats.effect.std.{Dispatcher, Semaphore}
 import cats.syntax.all._
 import cats.{Applicative, ApplicativeThrow, Functor, Show}
 import io.prometheus.client.Collector.MetricFamilySamples
-import io.prometheus.client.{
-  Collector,
-  CollectorRegistry,
-  CounterMetricFamily,
-  GaugeMetricFamily,
-  SimpleCollector,
-  SummaryMetricFamily,
-  Counter => PCounter,
-  Gauge => PGauge,
-  Histogram => PHistogram,
-  Info => PInfo,
-  Summary => PSummary
-}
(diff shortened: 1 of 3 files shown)


java/src/main/scala/prometheus4cats/javasimpleclient/internal/MetricCollectionProcessor.scala
```diff
@@ -209,84 +211,59 @@ class MetricCollectionProcessor[F[_]: Async: Logger] private (
...
-            "Because of this, only the first metric will be taken and the others ignored - this will not prevent your application from functioning.\n\n" +
-            "To mitigate this you can do the following to the metric collection:\n" +
-            "  - modify the metric name to disambiguate metric types and/or label sets\n" +
-            "  - align labels between the two metrics if they are of the same type, so that they both have the same keys,\n" +
-            "    but provide empty values for the labels which are not present on each metric"
-
         lazy val duplicatesInRegistry =
-          s"The following metrics with the same name and different labels or type are already registered in the metrics registry and therefore have been excluded from the provided collection: '${registryDuplicates
-              .mkString_(",")}'" + common +
+          s"The following metrics or callbacks with the same name are already registered in the metrics registry and therefore have been excluded from the provided collection: '${registryDuplicates
```
janstenpickle, line 255: I'm wonder whether we need a latch on this so as not to spam peoples' logs. Maybe only logging if the known duplicates changes? The problem is that it's _another_ ref to inspect when rendering `/metrics`

TimWSpence: Haha yeah, Honestly, we have so many refs floating around that I wouldn't worry too much about another

java/src/main/scala/prometheus4cats/javasimpleclient/JavaMetricRegistry.scala
```diff
@@ -63,8 +66,51 @@ class JavaMetricRegistry[F[_]: Async: Logger] private (
...
+          callbackTimeoutState,
+          stringName,
+          incTimeout,
+          Logger[F].warn(th)(
+            s"Timed out running callback for metric '$stringName' after $callbackTimeout.\n" +
+              "This may be due to a callback having been registered that performs some long running calculation which blocks\n" +
+              "Please review your code or raise an issue or pull request with the library from which this callback was registered.\n" +
+              s"This warning will only be shown once for each metric after process start. The counter '${JavaMetricRegistry.callbackErrorCounterName}'" +
+              "tracks the number of times this occurs per metric name."
+          ) >> incTimeout
```
TimWSpence, line 96: technically this should probably be `.guarantee`, although it's highly unlikely to make a difference

java/src/main/scala/prometheus4cats/javasimpleclient/JavaMetricRegistry.scala
```diff
@@ -63,8 +66,51 @@ class JavaMetricRegistry[F[_]: Async: Logger] private (
...
+  private def trackErrors[A](state: Ref[F, Set[String]], stringName: String, onContains: F[A], onContainsNot: F[A]) =
+    state.modify { current =>
+      if (current.contains(stringName)) (current, onContains) else (current + stringName, onContainsNot)
+    }.flatten
+
+  private def incErrorCounter(stringName: String, reason: String) =
+    Sync[F].delay(callbackErrorCounter.labels(stringName, reason).inc())
+
+  private def timeoutCallback[A](fa: F[A], empty: A, stringName: String): A = {
+    lazy val incTimeout = incErrorCounter(stringName, "timeout").as(empty)
```
TimWSpence, line 78: Interesting nitpick: this is only called once so you could also make it a `def` to avoid allocating if not required. Not sure how big the synchronization overhead of lazy is? Almost certainly it literally doesn't matter, just curious! 😂

TimWSpence (approved):
LGTM!

PR #63, opened 2023-03-16
Remove `name` as an illegal label name

janstenpickle:
Prometheus does not have this restriction, so safe to remove.


TimWSpence (approved):
LGTM!

PR #70, opened 2023-08-18
Get rid of `Sized` from APIs

alejandrohdezma:
## 🚀 What's included in this PR?

This PR removes the need to use `Sized` to check at compile time the size of parameters in the `labels` method. This is done by changing the method's signature from:

```scala
def labels[B, N <: Nat](labelNames: Sized[IndexedSeq[Label.Name], N])(
   f: B => Sized[IndexedSeq[String], N]
): LabelsBuildStep[F, A, B, N, L]
```

to...

```scala
def labels[B](labels: (Label.Name, B => String)*): LabelsBuildStep[F, A, B, L]
```


janstenpickle (approved):
gif

PR #71, opened 2023-08-18
Allow using `labels` the same way we use `label`

alejandrohdezma:
## 🚀 What's included in this PR?

This PR allows using `labels` interchangeably whenever we can use `label`, for both normal and "callback" DSLs. That means we can do this:

```scala
case class MyClass(a: String, b: Int)

factory
  .counter("test_total")
  .ofLong
  .help("some help")
  .label[String]("label1")
  .labels[MyClass](("a", _.a), ("b", _.b.toString()))
  .label[Int]("other_label")
  .build // Counter[F, Long, (String, MyClass, Int)]
```


janstenpickle (approved):
Aweome!

PR #106, opened 2025-07-29
Register HotSpot collectors by default

sideeffffect:


sideeffffect:
I've changed to not register the HotSpot collectors by default.

May I ask, why would doing so require a new major version?

Also a related question, why do you think somebody would prefer _not_ to have HotSpot collectors by default?

You will need to trigger the CI build, since I'm not a maintainer of this repository.

Thanks :pray:

alejandrohdezma:
Hey @sideeffffect, I don't know why somebody would prefer not to have it enabled. But start doing it after a patch or minor version feels wrong to me, that's why I prefer this being an opt-in feature.

PR #115, opened 2026-02-17
Reduce per-call allocation pressure in labelled metric operations

Oliver-Taylor:
## Summary

Every call to a labelled metric operation (`.inc()`, `.observe()`, etc.) currently performs several intermediate collection allocations that are avoidable:

1. **`labelNames ++ commonLabelNames`** — re-concatenates two `IndexedSeq`s on every call, despite the result being the same every time. Allocates a `VectorBuilder` and a new `Vector`.
2. **`f(labels) ++ commonLabelValues`** — concatenates dynamic label values with common label values via `IndexedSeq.++`. Allocates another `VectorBuilder` and `Vector`.
3. **`commonLabels.value.values.toArray`** — recomputed on every call despite being constant.
4. **`c.labels(indexedSeq: _*)`** — varargs expansion on an `IndexedSeq` calls `.toArray` internally, allocating a `String[]` via `Array.newInstance` + copy.
5. **`labelNames.flatMap(labels.get)`** in `unsafeLabels` — wraps each label lookup in `Option` (boxing/unboxing `Some`/`None` per element) then flattens via `VectorBuilder`.

In aggregate, each metric operation allocates ~3 `VectorBuilder`s, ~3 `Vector` results, N `Option` instances, and 1 `String[]` — all of which become immediate garbage.

### Pre-compute constant label data (`JavaMetricRegistry` + `Utils`)

- `labelNames ++ commonLabelNames` → pre-computed once as `allLabelNames` at metric construction time
- `commonLabels.value.values` → pre-computed once as `Array[String]` at metric construction time
- New `Utils.modifyMetric` overload accepts dynamic labels and pre-computed common label values separately
- New `buildLabelArray` method constructs the final `Array[String]` directly via a while loop + `System.arraycopy`, avoiding intermediate `IndexedSeq` concatenation
- The pre-built `Array[String]` is passed to `c.labels(arr: _*)`, which the Scala compiler passes directly without an additional `.toArray` copy

### Use `collect` instead of `flatMap` for label resolution (`package.scala`)

- Replace `labelNames.flatMap(labels.get)` with `labelNames.collect(labels)` in `unsafeLabels`
- `Map` extends `PartialFunction`, so `collect` preserves the existing filtering behaviour (missing keys are skipped) while avoiding `Option` boxing/unboxing per element

### Per-call allocation reduction

| Allocation | Before | After |
|---|---|---|
| `VectorBuilder` | 3 per call | 1 per call (`collect` result) |
| `Vector` / `IndexedSeq` results | 3 per call | 1 per call |
| `Option[String]` boxing | N per call (one per label) | 0 |
| `String[]` via `Array.newInstance` | 1 per call | 0 (direct `Array` pass) |
| `commonLabels.values.toArray` | 1 per call | 0 (pre-computed) |

All existing tests pass (43/43). No API changes — only internal implementation.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javasimpleclient/internal/Utils.scala
+++ b/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javasimpleclient/internal/Utils.scala
@@ -44,15 +44,15 @@ private[javasimpleclient] object Utils {
       logger(e)(s"Failed to unregister a collector: '$collector'")
     }
 
-  /** Builds a label array directly from dynamic label values and pre-computed common label values,
-    * avoiding intermediate IndexedSeq concatenation and varargs String[] allocation.
+  /** Builds a label array directly from dynamic label values and pre-computed common label values, avoiding
+    * intermediate IndexedSeq concatenation and varargs String[] allocation.
     */
   @inline private def buildLabelArray(
       dynamicLabels: IndexedSeq[String],
       commonLabelValues: Array[String]
   ): Array[String] = {
     val arr = new Array[String](dynamicLabels.length + commonLabelValues.length)
-    var i = 0
+    var i   = 0
     while (i < dynamicLabels.length) { arr(i) = dynamicLabels(i); i += 1 }
     System.arraycopy(commonLabelValues, 0, arr, dynamicLabels.length, commonLabelValues.length)
     arr
--- a/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javasimpleclient/internal/Utils.scala
+++ b/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javasimpleclient/internal/Utils.scala
@@ -52,8 +52,7 @@ private[javasimpleclient] object Utils {
       commonLabelValues: Array[String]
   ): Array[String] = {
     val arr = new Array[String](dynamicLabels.length + commonLabelValues.length)
-    var i   = 0
-    while (i < dynamicLabels.length) { arr(i) = dynamicLabels(i); i += 1 }
+    dynamicLabels.copyToArray(arr, 0): Unit
     System.arraycopy(commonLabelValues, 0, arr, dynamicLabels.length, commonLabelValues.length)
     arr
   }
--- a/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javasimpleclient/internal/Utils.scala
+++ b/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javasimpleclient/internal/Utils.scala
@@ -44,6 +44,20 @@ private[javasimpleclient] object Utils {
       logger(e)(s"Failed to unregister a collector: '$collector'")
     }
 
+  /** Builds a label array directly from dynamic label values and pre-computed common label values,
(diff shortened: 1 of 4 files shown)


alejandrohdezma (approved):

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javasimpleclient/internal/Utils.scala
```diff
@@ -78,11 +114,20 @@ private[javasimpleclient] object Utils {
       labelNames: IndexedSeq[Label.Name],
       labels: IndexedSeq[String]
   )(implicit F: Sync[F]): F[B] =
```
kidonm: **question: (implicits)** Is it intentional to pass implicits twice (once as `: Sync`, once as `implicit`)?

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javasimpleclient/internal/Utils.scala
```diff
@@ -78,11 +114,20 @@ private[javasimpleclient] object Utils {
...
-      child <- handlePrometheusCollectorErrors(
-                 F.delay(c.labels(labels: _*)), c, metricName, labelNames, labels
-               )
-    } yield child
+    handlePrometheusCollectorErrors(
+      F.delay(c.labels(labels: _*)), c, metricName, labelNames, labels
+    )
+
+  /** Overload that accepts a pre-built Array[String] to avoid varargs String[] allocation. */
+  private def retrieveCollectorForLabels[F[_], A: Show, B](
```
kidonm: **suggestion: (consistency)** This methods uses two approaches to declare its implicits. Consider unifying it.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javasimpleclient/internal/Utils.scala
```diff
@@ -98,6 +143,23 @@ private[javasimpleclient] object Utils {
...
     )
 
+  /** Overload that accepts Array[String] labels, only building the error map on failure. */
+  private def handlePrometheusCollectorErrors[F[_], A: Show, B](
+      fa: F[B],
+      c: SimpleCollector[_],
+      metricName: A,
+      labelNames: IndexedSeq[Label.Name],
+      labels: Array[String]
+  )(implicit F: Sync[F]): F[B] =
```
kidonm: **suggestion: (consistency)** This methods uses two approaches to declare its implicits. Consider unifying it.

kidonm (approved):
LGTM % style

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javasimpleclient/internal/Utils.scala
Oliver-Taylor: thanks, will sort out

kidonm (approved):

alejandrohdezma (approved):

alejandrohdezma (approved):

PR #123, opened 2026-04-30
v6: client_java 1.x backend with native histograms + NHCB-friendly dual-mode

gruffyn:
## Summary

Major-version uplift of the `prometheus4cats-java` adapter from the EOL
`io.prometheus:simpleclient` (0.16.0) line to the current
`io.prometheus:prometheus-metrics-core` (1.6.1) line. Adds **native
histogram** support and a **dual-mode (NHCB-friendly) histogram** path
that lets consumers preserve curated bucket boundaries while also
emitting native exponential.

> ⚠️ **Draft.** Sharing for early architectural review. The testkit suite
> wiring onto the new backend is in progress on a separate branch and
> will be folded in before merge — see "Not in this PR yet" below.

## Headline changes

- **New `prometheus4cats.javaclient.JavaMetricRegistry`** backed by
  `prometheus-metrics-core` 1.x. Same Builder-style construction API as
  the legacy `javasimpleclient.JavaMetricRegistry` (drop-in for
  consumers like `permutive-metrics`'s `PrometheusMetrics.scala:56`,
  modulo the package rename and a `PrometheusRegistry` parameter type
  in place of `CollectorRegistry`).
- **`NativeHistogram` config** + **`.nativeHistogram(...)` DSL builder**
  for native-exponential-only metrics.
- **`.histogram(...).withNative` (NHCB-friendly dual-mode)** — single
  declaration emits BOTH classic buckets AND native exponential.
  Designed as the pragmatic migration target for existing classic
  histograms with curated boundaries: classic survives for any
  Prometheus that doesn't speak native, native is available directly,
  and Prometheus 2.49+'s `convert_classic_histograms_to_nhcb` can
  perform server-side NHCB conversion on the classic representation.
  See commit `09b92ab` for the design rationale.
- **`Info` redesigned** to declare labels at build time
  (`.label[T]("...")` / `.labelsFrom[CaseClass]`), matching how every
  other metric type already works and matching the upstream 1.x Info
  API. v5→v6 source-level break for `Info[F, Map[Label.Name, String]]`
  consumers; wire format unchanged so alerts/dashboards/recording-rules
  are unaffected. See commit `66c907a` for migration shape.
- **Callback machinery** ported with per-callback timeout + combined
  timeout + log-once error tracking, matching legacy semantics.

## Commit-by-commit

13 commits, each green at every checkpoint (cross-build Scala 2.13.18 +
3.3.7, scalafmt + scalafix clean):

1. `129edec` — Add `NativeHistogram.Config`
[...]

PR #124, opened 2026-04-30
v6: client_java 1.x backend with native histograms + NHCB dual-mode

gruffyn:
## Summary

Major-version uplift of the `prometheus4cats-java` adapter from the EOL `io.prometheus:simpleclient` 0.x line to `io.prometheus:prometheus-metrics-core` 1.6.1. Adds per-histogram opt-in for native histograms and an NHCB-friendly dual-mode (`.histogram(...).withNative`).

Supersedes draft #123 (will be closed once this is merged). The history was replayed onto a fresh branch (`feat/v6-native-histograms`) so every commit carries a clean `Co-Authored-By` trailer matching the convention Oliver has used elsewhere.

## What changes for consumers

Source-incompatible (v6 = major bump, `versionPolicyIntention := Compatibility.None`):

- **Package rename**: `prometheus4cats.javasimpleclient` → `prometheus4cats.javaclient`. Anyone reaching past the abstraction (e.g. instantiating `JavaMetricRegistry.Builder` directly) needs the import update. Builder API surface is unchanged.
- **Underlying registry type**: `io.prometheus.client.CollectorRegistry` → `io.prometheus.metrics.model.registry.PrometheusRegistry`. Consumers passing one in via `.withRegistry(...)` need the type swap.
- **`Summary.Value.count: Double` → `Long`**. Counts are integer in Prometheus; v5 typed them as the value-type parameter `A` for parametricity, v6 fixes the type to match the underlying contract and the new `SummaryDataPointSnapshot` constructor. Compile errors at every call site (no silent breaks); fix is `.toLong` insertion or `1.0` literal → `1L`.
- **`Info` declares labels at build time, not at observation time**. Was a v5 quirk that didn't fit the upstream 1.x model.

## Native histogram API

Three entrypoints, each maps to one upstream builder configuration:

| Entrypoint | Classic emission | Native emission |
|---|---|---|
| `factory.histogram(...).buckets(...)` | ✓ (default) | ✗ |
| `factory.histogram(...).buckets(...).withNative` | ✓ (NHCB-friendly dual mode) | ✓ |
| `factory.nativeHistogram(...)` | ✗ | ✓ (exponential-only) |

`withNative` is the recommended migration path for existing classic histograms with curated bucket boundaries: preserves bucket intent for any Prometheus that doesn't speak native, emits a native exponential alongside, and Prometheus 2.49+'s `convert_classic_histograms_to_nhcb` can convert classic→NHCB at scrape time.

`.nativeHistogram(...)` is the right choice for new metrics that don't have bucket intent worth preserving.

## ⚠️ Open question for review (Alex)

The v5 javasimpleclient registry exposed a set of self-observability metrics — `prometheus4cats_registered_metrics`, `prometheus4cats_registered_metric_claims`, `prometheus4cats_registered_callback_metrics`, `prometheus4cats_registered_callbacks_per_metric`, `prometheus4cats_combined_callback_metric_total`, `prometheus4cats_callback_total`, `prometheus4cats_collection_callback_duplicates`, `prometheus4cats_collection_callback_duration`, `prometheus4cats_combined_collection_callback_total`, `prometheus4cats_collection_callback_total`. The v6 backend currently does NOT emit them.

We don't have any internal usage of these metrics, so we are wondering if there's value in keeping them. The docs explicitly advertised them as a public contract, so silently dropping them is the kind of regression that lands wrong with downstream maintainers — but if no-one's using them, the simplest thing is to document the removal and move on.

I marked the section in `website/docs/implementations/java.md` as "v6 regression — pending" so the docs don't lie about what the registry emits.

**Three options to choose from:**

1. **Re-add all v5 self-metrics before v6.0.0 is tagged.** ~150-250 LoC across `JavaMetricRegistry.scala` (constructor + builder + the existing callback flow). The state already exists internally; we'd just be exposing it. Maximum compatibility, minimum surprise.
2. **Re-add a subset** (the gauges that just observe state — cheap; skip the timing histogram if it was rarely used).
3. **Drop them entirely**, document the breakage in the migration guide, and let consumers re-derive what they need from `/metrics` directly.

I'd lean (1) — for a major version we get to break source/binary compat, but breaking the wire-format names of self-metrics consumers run dashboards on for no good reason is gratuitous when the legacy implementation already gave us the design. Want your call.

## Test results

Cross-built on Scala 2.13.18 + 3.3.7. All green:

| Suite | Tests |
|---|---|
| `prometheus4cats` (core) | 42 |
| `prometheus4cats-testing` | 25 |
| `prometheus4cats-java` (`JavaMetricRegistryTestkitSuite` + `JavaMetricRegistrySuite`) | 57 (38 testkit + 19 JVM-specific) |

The 57 javaclient tests include 5 name-collision contract tests promoted from the deleted legacy suite, which uncovered two real backend bugs during the uplift (cross-axis metric↔callback collision detection was missing; the metric-vs-callback registration paths keyed under different `_total`-stripped vs full names so same-name detection missed counter cases). Both fixed in the testkit-conformance commit.

## What's deferred

- **Migration guide** (`website/docs/migration-from-v5.md` and `website/docs/native-histograms.md`) — Workstreams 5a and 5b in the plan. Worth a follow-up PR.
- **`prometheus4cats-contrib` v6 bump** — separate repo, separate PR.
- **`permutive-metrics` bridge** + **`scala-bom`** rollout — internal repos, separate PRs.
- **Self-metrics** — see open question above; pending decision.

## Test plan

- [ ] Cross-build green on CI (Scala 2.13 + 3.3)
- [ ] mdoc compiles all docs
- [ ] `versionPolicyCheck` passes against `Compatibility.None`
- [ ] Canary on `reference-app-scala` end-to-end (publishLocal + scrape verification) before tagging v6.0.0


gruffyn:
Additional data point on the self-metrics question (the "⚠️ Open question for review" section in the PR body):

While doing the lockstep `prometheus4cats-contrib` v6 bump (separate PR coming), I found one additional usage of `prometheus4cats_collection_callback_duplicates` — but it's a **diagnostic probe in a test**, not a real consumer reliance. The fs2-kafka contrib's `KafkaMetricsSuite` was asserting that the metric existed with `.samples.isEmpty` to verify no duplicate metric registrations slipped through during test setup.

That probe is now redundant in v6: the new javaclient backend's collision detection (including the cross-axis metric↔callback collision check landed in the testkit-conformance commit on this PR) raises `IllegalArgumentException` at registration time, so duplicates can't slip through silently — they crash the registration eagerly. I've dropped the assertions in the contrib v6 PR and the test invariant is preserved by the stricter v6 behaviour.

Net rollup so far across the org and codebase:
- Production Grafana / prometheus-rules: 0 references
- prometheus4cats GitHub issues + PRs: 0 references
- Source-code consumers (org-wide gh code search, archived excluded): 0
- prometheus4cats-contrib test diagnostic use: 1, replaceable by v6's stricter collision throw

If anything this strengthens the case for the "drop them, document the breakage in the migration guide" option (option 3) rather than the "re-add to be safe" option (option 1).

PR #128, opened 2026-05-19
Add NativeHistogram config type for native histogram tuning

hgdyeo:
Adds the `prometheus4cats.NativeHistogram` config case class — tuning parameters that subsequent v6 PRs will pass when registering native histograms.

## Summary

- New file `modules/prometheus4cats/src/main/scala/prometheus4cats/NativeHistogram.scala`
- Five fields: `initialSchema`, `maxNumberOfBuckets`, `resetDuration`, `minZeroThreshold`, `maxZeroThreshold`
- Wither pattern matching every other prometheus4cats tuning type
- `NativeHistogram.Default` matches upstream Java client defaults (schema=8, maxBuckets=160, no reset, no zero growth)

## Context

PR 2 of the v6 backend uplift split out of #124. Stacked on top of #127 (deps + version policy).

The type is unused at this point — `MetricRegistry` and `MetricFactory` gain native histogram methods in the next PR (3), and the new `javaclient` adapter wires it up to the upstream `Histogram.Builder` later (PR 7). Reviewing this PR independently is reviewing the public type shape on its merits before any code references it.

## Why this PR is split out

Locking the public tuning surface ahead of any consumer of it lets reviewers focus on field naming, defaults, and the wither-pattern shape without distraction. The whole change is one 74-line file.

## Risk

Zero. Pure addition; the type has no consumers in this PR.

## Test plan

- [x] `sbt fix --check` clean

## Reviewer focus

- Are the defaults sensible? (Verified upstream: schema range `[-4, 8]` with 8 = max resolution; maxBuckets default 160 matches upstream.)
- Field naming maps to upstream `Histogram.Builder` setters (`nativeInitialSchema`, `nativeMaxNumberOfBuckets`, etc.) — translation in PR 7 will be mechanical.
- `final case class NativeHistogram private (…)` — note that `private` here only makes `apply` private; `copy` is still public, so consumers can bypass the wither pattern via `Default.copy(...)`. Worth deciding whether to tighten this.
- No input validation on withers — upstream rejects schema outside `[-4, 8]` at registration time. Worth catching at construction.

### DIFF
--- a/modules/prometheus4cats/src/main/scala/prometheus4cats/NativeHistogram.scala
+++ b/modules/prometheus4cats/src/main/scala/prometheus4cats/NativeHistogram.scala
@@ -29,8 +29,9 @@ import scala.concurrent.duration._
   *
   * @param initialSchema
   *   The schema controls native histogram resolution. Each higher schema halves bucket width (roughly: schema `s`
-  *   produces `2^s` buckets per power of two of the observation range). Valid range is `[-4, 8]`. Default `8` matches
-  *   upstream and gives the highest available resolution.
+  *   produces `2^s` buckets per power of two of the observation range). Valid range is `[-4, 8]`. Default `5` matches
+  *   upstream — chosen as a production sweet-spot that provides enough resolution without exceeding
+  *   `maxNumberOfBuckets` on typical workloads.
   * @param maxNumberOfBuckets
   *   Upper limit on the number of populated native buckets. When exceeded, the histogram automatically reduces its
   *   schema. Default `160`. Set lower to bound per-metric memory; set to `0` to disable the limit.
@@ -38,10 +39,11 @@ import scala.concurrent.duration._
   *   How often the histogram resets back to the original schema. Useful when transient observations push the schema
   *   down and you want it to recover. Default `0.seconds` (no reset).
   * @param minZeroThreshold
-  *   Smallest tracked zero-bucket threshold. Observations below this are folded into the zero bucket. Default `0.0`.
+  *   Smallest tracked zero-bucket threshold. Observations below this are folded into the zero bucket. Default `2^-128`
+  *   (≈ `2.94e-39`) matches upstream and provides a sub-denormal safety floor.
   * @param maxZeroThreshold
   *   Largest allowed zero-bucket threshold. The histogram may grow the zero bucket up to this value when the
-  *   `maxNumberOfBuckets` limit is hit. Default `0.0` (no growth).
+  *   `maxNumberOfBuckets` limit is hit. Default `2^-128` (≈ `2.94e-39`) matches upstream.
   */
 final case class NativeHistogram private (
     initialSchema: Int,
@@ -67,8 +69,11 @@ object NativeHistogram {
 
   /** Default tuning matching the upstream Java client defaults. */
   val Default: NativeHistogram = new NativeHistogram(
-    initialSchema = 8, maxNumberOfBuckets = 160, resetDuration = 0.seconds, minZeroThreshold = 0.0,
-    maxZeroThreshold = 0.0
+    initialSchema = 5,
+    maxNumberOfBuckets = 160,
+    resetDuration = 0.seconds,
+    minZeroThreshold = math.pow(2.0, -128),
+    maxZeroThreshold = math.pow(2.0, -128)
   )
 
 }
--- a/modules/prometheus4cats/src/main/scala/prometheus4cats/NativeHistogram.scala
+++ b/modules/prometheus4cats/src/main/scala/prometheus4cats/NativeHistogram.scala
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2022-2026 Permutive Ltd. <https://permutive.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


modules/prometheus4cats/src/main/scala/prometheus4cats/NativeHistogram.scala
```diff
@@ -0,0 +1,74 @@
...
+
+  def withMaxZeroThreshold(maxZeroThreshold: Double): NativeHistogram = copy(maxZeroThreshold = maxZeroThreshold)
+
+}
+
+object NativeHistogram {
+
+  /** Default tuning matching the upstream Java client defaults. */
+  val Default: NativeHistogram = new NativeHistogram(
+    initialSchema = 8, maxNumberOfBuckets = 160, resetDuration = 0.seconds, minZeroThreshold = 0.0,
```
hgdyeo: **`initialSchema = 8` does not match upstream's default of `5`.**

Upstream's `Histogram.java` defines:

```java
private static final int DEFAULT_NATIVE_INITIAL_SCHEMA = 5;
```

This matters more than it might look:

- Schema 8 produces **256 buckets per power-of-two** of observation range. Schema 5 produces **32**. That's an 8× difference in granularity per range decade.
- With `maxNumberOfBuckets = 160` (which we *do* match), schema 8 will exceed the bucket cap on most realistic workloads almost immediately — triggering upstream's automatic schema reduction. The histogram then runs at a coarser effective schema than 8 anyway, defeating the higher initial setting and burning CPU on reductions.
- Schema 5 was chosen upstream as the production-sweet-spot: enough resolution for observability without hitting the cap. The Go and Python Prometheus clients also default to 5.

The scaladoc on this `val` (line 68 above) says `"Default tuning matching the upstream Java client defaults"` — and the field scaladoc at line 30 says `"Default 8 matches upstream and gives the highest available resolution"`. Both are inaccurate for this field.

**Recommended fix:** change to `initialSchema = 5` and correct the scaladoc.

hgdyeo: **`minZeroThreshold = 0.0` does not match upstream's default of `2^-128` (≈ `2.94e-39`).**

Upstream's `Histogram.java`:

```java
private static final double DEFAULT_NATIVE_MIN_ZERO_THRESHOLD = Math.pow(2.0, -128);
```

For typical workloads (latencies, byte counts) `0.0` vs `2^-128` is indistinguishable in practice — both effectively mean "no observation folds into the zero bucket". But:

- `0.0` literally means observations of `+0.0` and `-0.0` go to the zero bucket and nothing else does.
- `2^-128` (~`2.94e-39`, sub-denormal) provides a tiny safety floor so absurdly small observations don't inflate bucket counts.

Less load-bearing than the schema mismatch, but the scaladoc says we match upstream — we don't. Match upstream unless there's a specific reason not to.

modules/prometheus4cats/src/main/scala/prometheus4cats/NativeHistogram.scala
```diff
@@ -0,0 +1,74 @@
...
+    maxZeroThreshold = 0.0
```
hgdyeo: **`maxZeroThreshold = 0.0` does not match upstream's default of `2^-128` (≈ `2.94e-39`).**

Upstream's `Histogram.java`:

```java
private static final double DEFAULT_NATIVE_MAX_ZERO_THRESHOLD = Math.pow(2.0, -128);
```

This one is slightly more consequential than `minZeroThreshold`: `maxZeroThreshold` caps how far the histogram can adaptively grow the zero bucket when the `maxNumberOfBuckets` limit is hit. Setting `0.0` means the zero bucket cannot grow at all — so under bucket pressure, upstream falls back to schema reduction instead of zero-bucket widening. That changes the bucket-pressure-relief strategy.

Note also that the upstream API doc for `nativeMaxZeroThreshold` has a typo claiming the default is `DEFAULT_NATIVE_MAX_NUMBER_OF_BUCKETS` (the bucket-count default, 160) — that's clearly wrong. The actual source value is `2^-128`.

Match upstream's `Math.pow(2.0, -128)`.

hgdyeo (commented):
Three discrepancies between `NativeHistogram.Default` and upstream `prometheus-metrics-core` 1.x defaults. The scaladoc and commit body both claim the defaults match upstream — they don't. Verified against [Histogram.java in client_java](https://github.com/prometheus/client_java/blob/main/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Histogram.java).

All three inline below.

Otherwise, the interface aligns with upstream and the method names are sensible

modules/prometheus4cats/src/main/scala/prometheus4cats/NativeHistogram.scala
alejandrohdezma: Verified against `Histogram.java:699` (`DEFAULT_NATIVE_INITIAL_SCHEMA = 5`). Fixed in 832ae76 along with the scaladoc — also tightened the field doc to mention the production-sweet-spot rationale rather than the resolution claim.

alejandrohdezma: Confirmed against `Histogram.java:697` (`Math.pow(2.0, -128)`). Fixed in 832ae76, scaladoc updated to match.

modules/prometheus4cats/src/main/scala/prometheus4cats/NativeHistogram.scala
alejandrohdezma: Confirmed against `Histogram.java:698`. Fixed in 832ae76. Side note on the upstream doc typo (`Histogram.java:917` still says `@link Builder#DEFAULT_NATIVE_MAX_NUMBER_OF_BUCKETS` on the `nativeMaxZeroThreshold` setter) — worth filing upstream when we get a chance.

alejandrohdezma (approved):

PR #129, opened 2026-05-19
Extend MetricRegistry SPI and MetricFactory DSL for native histograms

hgdyeo:
Wires native histograms into the `MetricRegistry` SPI and `MetricFactory` DSL. Source-incompatible (intended for v6 — adds an abstract method to a public trait).

## Summary

- `MetricRegistry.createAndRegisterDoubleNativeHistogram[A]` — new abstract method, no Long variant (see reviewer notes)
- noop and `mapK` implementations delegate cleanly
- `MetricFactory.nativeHistogram(name, config = NativeHistogram.Default)` — returns `HelpStep[MetricDsl[F, Double, Histogram]]`, deliberately skipping the `.ofLong/.ofDouble` step since native histograms are inherently double-valued
- Stubs in `TestingMetricRegistry` (records observations with `ExemplarState.noop`), the legacy `javasimpleclient` adapter (raises `UnsupportedOperationException` pointing to v6+), and `ExternalPackageMetricRegistry` (proves external impls can satisfy the SPI)
- DSL coverage in `MetricsFactoryDslTest`: the new `.nativeHistogram(...)` chain compiles end-to-end including `.asTimer`, `.label[…]`, `.unsafeLabels`, and the explicit-config overload

## Example

```scala
metrics.nativeHistogram("my_histogram").help("...").label[Int]("first_label").build

// with custom tuning:
metrics
  .nativeHistogram("my_histogram", NativeHistogram.Default.withInitialSchema(6))
  .help("...")
  .label[Int]("first_label")
  .build
```

## Context

PR 3 of the v6 backend uplift split out of #124. Stacked on #128 (NativeHistogram type).

## Why this PR is split out

This is the first source-breaking change in the v6 series (adds an abstract method to a public trait — external implementers like `prometheus4cats-contrib` will need to update). Isolating it as a single PR lets reviewers focus on SPI shape decisions before the new backend starts referencing them in PR 6+.

## Risk

Source-incompatible for external `MetricRegistry` implementers — intended. The `ExternalPackageMetricRegistry` test fixture in this PR proves the new method can be satisfied from outside the library.

## Test plan

- [x] `sbt fix --check` clean

## Reviewer focus

- **No Long variant.** The commit body argues "native histograms are inherently double-valued — convert at the call site". But PR 5's dual-mode method has both Double and Long variants. Asymmetric. Should be reconciled — probably by adding a Long variant here too (contramap-derived in `DoubleMetricRegistry`).
- **No callback variant.** Consumers wanting callback-driven native histograms can't have them. Worth a scaladoc breadcrumb on `MetricFactory.WithCallbacks.histogram` pointing to this gap.
- **`nativeHistogram: NativeHistogram` parameter name collides with the type name** throughout the diff. Cosmetic but pervasive — worth a rename to `config` or `tuning`.
- **DSL exit type `HelpStep[MetricDsl[F, Double, Histogram]]` skips the type step.** Justified — but the asymmetry vs `.histogram(...).ofLong / .ofDouble` should be one bullet in the v6 migration guide (PR 11).

### DIFF
--- a/modules/prometheus4cats-testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
+++ b/modules/prometheus4cats-testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
@@ -46,7 +46,9 @@ sealed abstract class TestingMetricRegistry[F[_]] private (
         )
       ]
     ],
-    private val info: MapRef[F, String, Option[(Int, Info[F, Map[Label.Name, String]])]]
+    // Storage for Info reference counts. Tracks "is this info name claimed", not the Info instance
+    // itself, so we don't have to encode the per-call labels-type parameter A in this map.
+    private val info: MapRef[F, String, Option[Int]]
 )(implicit override val F: Concurrent[F])
     extends DoubleMetricRegistry[F]
     with DoubleCallbackRegistry[F] {
@@ -385,24 +387,23 @@ sealed abstract class TestingMetricRegistry[F[_]] private (
       Chain.nil
     )
 
-  override def createAndRegisterInfo(
+  override def createAndRegisterInfo[A](
       prefix: Option[Metric.Prefix],
       name: Info.Name,
-      help: Metric.Help
-  ): Resource[F, Info[F, Map[Label.Name, String]]] = {
-    val release = info(NameUtils.makeName(prefix, name.value)).update {
-      case None         => throw new RuntimeException("This should be unreachable - our reference counting has a bug")
-      case Some((n, i)) => if (n == 1) None else Some(n - 1 -> i)
-
+      help: Metric.Help,
+      labelNames: IndexedSeq[Label.Name]
+  )(f: A => IndexedSeq[String]): Resource[F, Info[F, A]] = {
+    val key = NameUtils.makeName(prefix, name.value)
+    val release = info(key).update {
+      case None    => throw new RuntimeException("This should be unreachable - our reference counting has a bug")
+      case Some(n) => if (n == 1) None else Some(n - 1)
     }
-    Resource.make(
-      info(NameUtils.makeName(prefix, name.value)).modify {
-        case None =>
-          val i = Info.make[F, Map[Label.Name, String]]((_: Map[Label.Name, String]) => F.unit)
-          Some(1 -> i) -> i
-        case Some((n, i)) => Some(n + 1 -> i) -> i
-      }
-    )(_ => release)
+    Resource
+      .make(info(key).update {
+        case None    => Some(1)
+        case Some(n) => Some(n + 1)
+      })(_ => release)
+      .as(Info.make[F, A]((_: A) => F.unit))
   }
 
   override def registerDoubleCounterCallback[A](
@@ -537,7 +538,7 @@ object TestingMetricRegistry {
             Ref[F, Option[Exemplar.Data]]
         )
       ](256),
-    MapRef.ofShardedImmutableMap[F, String, (Int, Info[F, Map[Label.Name, String]])](64)
+    MapRef.ofShardedImmutableMap[F, String, Int](64)
   ).mapN { case (m, i) => new TestingMetricRegistry(m, i) {} }
 
   sealed trait MetricType
--- a/modules/prometheus4cats-testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
+++ b/modules/prometheus4cats-testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
@@ -352,7 +352,7 @@ sealed abstract class TestingMetricRegistry[F[_]] private (
       help: Metric.Help,
       commonLabels: Metric.CommonLabels,
       labelNames: IndexedSeq[Label.Name],
-      nativeHistogram: NativeHistogram
+      config: NativeHistogram
   )(f: A => IndexedSeq[String]): Resource[F, Histogram[F, Double, A]] =
     store(
       NameUtils.makeName(prefix, name.value),
--- a/modules/prometheus4cats-testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
+++ b/modules/prometheus4cats-testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
@@ -368,6 +368,27 @@ sealed abstract class TestingMetricRegistry[F[_]] private (
       Chain.nil
     )
 
+  override def createAndRegisterDoubleHistogramWithNative[A](
+      prefix: Option[Metric.Prefix],
+      name: Histogram.Name,
+      help: Metric.Help,
+      commonLabels: Metric.CommonLabels,
+      labelNames: IndexedSeq[Label.Name],
+      buckets: NonEmptySeq[Double],
+      nativeHistogram: NativeHistogram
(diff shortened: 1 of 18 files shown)


modules/prometheus4cats/src/main/scala/prometheus4cats/MetricRegistry.scala
```diff
@@ -205,6 +205,43 @@ trait MetricRegistry[F[_]] {
...
+    * @return
+    *   a [[Histogram]] wrapped in whatever side effect that was performed in registering it
+    */
+  def createAndRegisterDoubleNativeHistogram[A](
+      prefix: Option[Metric.Prefix],
+      name: Histogram.Name,
+      help: Metric.Help,
+      commonLabels: Metric.CommonLabels,
+      labelNames: IndexedSeq[Label.Name],
+      nativeHistogram: NativeHistogram
```
hgdyeo: Nit — the SPI parameter `nativeHistogram` shares its name with its type, which reads awkwardly at the call site:

```scala
createAndRegisterDoubleNativeHistogram(..., nativeHistogram = NativeHistogram.Default)
```

The DSL layer already uses `config: NativeHistogram` ([`MetricFactory.scala:234`](https://github.com/permutive-engineering/prometheus4cats/blob/feat/v6-native-histogram-spi/modules/prometheus4cats/src/main/scala/prometheus4cats/MetricFactory.scala#L234)). Worth aligning across the 6 SPI sites:

- `MetricRegistry.scala:242` (abstract)
- `MetricRegistry.scala:388` (noop)
- `MetricRegistry.scala:471` (mapK)
- `TestingMetricRegistry.scala:355`
- `javasimpleclient/JavaMetricRegistry.scala:311`
- `ExternalPackageMetricRegistry.scala:94`

Happy with `config`, `tuning`, or anything that isn't the type name.

modules/prometheus4cats/src/main/scala/prometheus4cats/MetricRegistry.scala
```diff
@@ -205,6 +205,43 @@ trait MetricRegistry[F[_]] {
...
+  )(f: A => IndexedSeq[String]): Resource[F, Histogram[F, Double, A]]
```
hgdyeo, line 280: ## Long-variant asymmetry — worth a maintainer discussion

Quick mental map of the three histogram modes and their Long variants:

| Mode | DSL | SPI (Double) | SPI (Long) |
|---|---|---|---|
| Classic-only | `.histogram(...).buckets(...)` | ✓ | ✓ |
| Native-only (this PR) | `.nativeHistogram(...)` | ✓ | **missing** |
| Dual-mode (PR #131) | `.histogram(...).buckets(...).withNative` | ✓ | ✓ |

Modes A and C let consumers observe `Long`-typed values via a contramap derivation in `DoubleMetricRegistry`. Mode B (this PR) forces `.toDouble` at every observation site — purely because the convenience method isn't there, not anything fundamental about native histograms.

(Sanity-checked: the Long variant isn't added later in #124's history either, so this isn't something the split lost — it's a genuine asymmetry in the original work.)

Two coherent stances, either closes this:

**(a) Keep Double-only, document it.** The call-site `.toDouble` is treated as a feature — it makes consumers acknowledge that native histograms operate in real-valued space rather than letting them sail through with muscle memory from classic. Action: add a `@note` here on the abstract method and a corresponding line on the `.nativeHistogram(...)` DSL scaladoc saying "Long variant intentionally omitted — convert at the call site".

**(b) Add the Long variant.** Treat Long-input as the same call-site ergonomic we offer everywhere else in the library. Action: add `createAndRegisterLongNativeHistogram` to `DoubleMetricRegistry.scala` (contramap-derived, ~10 lines like the existing classic Long pattern), and wrap the DSL in `TypeStep` so `.nativeHistogram(...).ofLong / .ofDouble` works.

No strong preference from us — just want the asymmetry to be a deliberate call rather than an accidental one. Maintainer call.

hgdyeo (commented):
Two comments — one nit on parameter naming, one open question about Long-variant symmetry. Both inline below.

modules/prometheus4cats/src/main/scala/prometheus4cats/MetricRegistry.scala
hgdyeo: I'm going to go with (a) for now as it's less changed code from the original, but we can always change this.

modules/prometheus4cats/src/main/scala/prometheus4cats/MetricRegistry.scala
hgdyeo: I will change to `config`

modules/prometheus4cats/src/main/scala/prometheus4cats/MetricRegistry.scala
kidonm: could we chat about this offline before merging?

The reasoning here ("buckets are real-valued, so Double-only") makes sense to me, but I noticed classic histograms keep the `ofLong` / `ofDouble` split even though `Long` is really sugar over `Double` at the backend — `DoubleMetricRegistry` converts both bucket boundaries and observations to `Double` before they reach Prometheus. I wonder if one unstated benefit of that split is steering users away from configuring count-shaped histograms with `Double` boundaries — the type acting as a soft guardrail, not just ergonomic sugar. If that is part of why the split exists, dropping `.ofLong` for native histograms loses it. Worth a quick conversation outside GitHub to make sure we are aligned.

modules/prometheus4cats-testing/src/main/scala/prometheus4cats/testing/TestingMetricRegistry.scala
```diff
@@ -346,6 +346,26 @@ sealed abstract class TestingMetricRegistry[F[_]] private (
...
+      name: Histogram.Name,
+      help: Metric.Help,
+      commonLabels: Metric.CommonLabels,
+      labelNames: IndexedSeq[Label.Name],
+      config: NativeHistogram
+  )(f: A => IndexedSeq[String]): Resource[F, Histogram[F, Double, A]] =
+    store(
+      NameUtils.makeName(prefix, name.value),
+      names(commonLabels, labelNames),
+      MetricType.Histogram,
```
kidonm, line 362: **question** (non-blocking): how does this behave if a test registers a classic histogram and a native histogram under the same name?

Both end up tagged as `MetricType.Histogram` here, so I am not sure whether `store` would treat them as the same metric, raise a duplicate-registration error, or something else — and I could not tell from the surrounding code which is intended. I had a peek through the stack (#130 → #133) and noticed the new `javaclient` backend in #132 adds a dedicated `MetricType.NativeHistogram` to its private models, but `TestingMetricRegistry` does not gain a matching variant anywhere in the stack — at the top of #133 both `createAndRegisterDoubleNativeHistogram` and `createAndRegisterDoubleHistogramWithNative` still use `MetricType.Histogram`. Curious whether that is deliberate (because they really are interchangeable from a test-assertion standpoint) or just has not come up yet.

kidonm (commented):
Code review with 2 comments, both non-blocking.

modules/prometheus4cats/src/main/scala/prometheus4cats/MetricRegistry.scala
hgdyeo: I hadn't considered this to be honest, it's a good point. I honestly don't know if that's part of why the split exists.

PR #130, opened 2026-05-19
Redesign Info to declare labels at build time (v5->v6 API break)

hgdyeo:
Aligns `Info` with how every other metric type in this library already works: labels declared at build time using `.label[T](...)`, observed via positional values, instead of `Info[F, Map[Label.Name, String]]` accepting a Map at every `info.info(...)` call.

**This change is forced by upstream.** `prometheus-metrics-core` 1.x's `Info.builder()` declares `labelNames(...)` at build time and exposes only positional `setLabelValues(...)` at observation — there is no API to register an info with map-style labels at runtime.

## Before / after

```scala
// v5
factory.info("build_info").help("...").build  // Info[F, Map[Label.Name, String]]
i.info(Map(Label.Name("version") -> "1.0", Label.Name("commit") -> "abc"))

// v6
factory.info("build_info").help("...")
  .label[String]("version")
  .label[String]("commit").build       // Info[F, (String, String)]
i.info(("1.0", "abc"))

// or via Label.Encoder:
factory.info("build_info").help("...").labelsFrom[BuildInfo].build
i.info(buildInfo)
```

Wire format on the Prometheus side is **unchanged** (`build_info{version="…",commit="…"} 1`). Purely a Scala-side API break — every Info call site changes (compile errors at all of them, no silent breaks).

## Summary

- `MetricRegistry.createAndRegisterInfo[A]` now takes `labelNames: IndexedSeq[Label.Name]` and a curried `f: A => IndexedSeq[String]` projection, matching every other `createAndRegister` method on the trait.
- `MetricFactory.info(name)` returns `HelpStep[MetricDsl[F, Unit, InfoL]]` where `InfoL[F[_], A, B] = Info[F, B]` (a new type alias in the package object). This reuses the existing `MetricDsl` / `LabelledMetricDsl` machinery rather than maintaining a parallel mini-DSL.
- `javasimpleclient` adapter still passes `IndexedSeq.empty` to the underlying simpleclient builder and constructs the legacy `info(Map<String, String>)` call from `labelNames` + values at observation time. Preserves v5 wire format.
- `TestingMetricRegistry` simplified — drops the cached Info instance per name and just tracks refcounts.
- The new `javaclient` adapter Info implementation lands in PR 8, alongside Summary (the `javaclient` package doesn't exist on this branch yet).

## Context

PR 4 of the v6 backend uplift split out of #124. Stacked on #129 (native histogram SPI).

## Why this PR is split out

The Info redesign is conceptually independent of native histograms. Reviewing it in isolation lets reviewers focus on the API shape and the simpleclient-adapter rework without distraction from histogram work.

## Risk

Every `Info` call site downstream changes. Compile errors at every call site (no silent breaks); fix is mechanical (`.label[T](name)` declarations + positional observe).

## Test plan

- [x] `sbt fix --check` clean

## Reviewer focus

- **`InfoL[F[_], A, B] = Info[F, B]` type alias** — one-line workaround to fit `Info` (two type params) into the existing `MetricDsl` machinery (parameterised over three). The unused `A` is documented but it's still a kind-shape mismatch future maintainers will notice. Decision: accept the alias, or refactor `MetricDsl` to support both shapes (significantly more invasive).
- **`InfoMapSyntax` extension class still exists on `Info` companion.** Anyone with an `Info[F, Map[Label.Name, String]]` (e.g. via `mapK`) keeps the `.info(Map)` ergonomic. Decide whether this is intentional (back-compat ergonomics) or oversight.
- **`Info[F, Unit]` (no labels declared)** — upstream 1.x's `setLabelValues()` on an info with empty `labelNames` emits no data point. v5 callers using `factory.info(name).help(h).build` as an existence marker will emit nothing on the v6 javaclient backend. Migration-guide item; worth a scaladoc note on `MetricFactory.info` too.
- **`TestingMetricRegistry` behaviour change**: no longer returns the same cached `Info` instance across nested registrations of the same name. Test suite updated; downstream code using `TestingMetricRegistry` directly should be aware.

### DIFF
--- a/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
+++ b/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
@@ -281,6 +281,70 @@ class JavaMetricRegistry[F[_]: Async] private (
     }
   }
 
+  override def createAndRegisterDoubleHistogramWithNative[A](
+      prefix: Option[Metric.Prefix],
+      name: Histogram.Name,
+      help: Metric.Help,
+      commonLabels: Metric.CommonLabels,
+      labelNames: IndexedSeq[Label.Name],
+      buckets: NonEmptySeq[Double],
+      nativeHistogram: NativeHistogram
+  )(f: A => IndexedSeq[String]): Resource[F, Histogram[F, Double, A]] = {
+    val commonLabelNames       = commonLabels.value.keys.toIndexedSeq
+    val commonLabelValuesArray = commonLabels.value.values.toArray
+    val allLabelNames          = labelNames ++ commonLabelNames
+    val fullName               = NameUtils.makeName(prefix, name)
+
+    configureBuilderOrRetrieve[PHistogram](
+      register = () => {
+        // Dual-mode: NEITHER .classicOnly() NOR .nativeOnly(). Both classicUpperBounds(...) and
+        // nativeInitialSchema(...) are set. The resulting Histogram emits BOTH representations,
+        // letting Prometheus's `convert_classic_histograms_to_nhcb` pick the classic form for
+        // server-side NHCB conversion while the native exponential is also available directly.
+        val builder = PHistogram
+          .builder()
+          .name(fullName)
+          .help(help.value)
+          .labelNames(allLabelNames.map(_.value): _*)
+          .classicUpperBounds(buckets.toSeq.toArray: _*)
+          .nativeInitialSchema(nativeHistogram.initialSchema)
+          .nativeMaxNumberOfBuckets(nativeHistogram.maxNumberOfBuckets)
+          .nativeMaxZeroThreshold(nativeHistogram.maxZeroThreshold)
+          .nativeMinZeroThreshold(nativeHistogram.minZeroThreshold)
+        val tuned =
+          if (nativeHistogram.resetDuration > 0.seconds)
+            builder.nativeResetDuration(
+              nativeHistogram.resetDuration.toSeconds,
+              java.util.concurrent.TimeUnit.SECONDS
+            )
+          else builder
+        tuned.register(registry)
+      },
+      // Use a distinct MetricType so dedup is correct: registering the same metric name as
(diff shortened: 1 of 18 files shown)


hgdyeo:
**Recommend deleting `InfoMapSyntax`** in this PR.

Location: [`modules/prometheus4cats/src/main/scala/prometheus4cats/Info.scala:59-62`](https://github.com/permutive-engineering/prometheus4cats/blob/feat/v6-info-redesign/modules/prometheus4cats/src/main/scala/prometheus4cats/Info.scala#L59-L62) (file not modified by this PR but the extension class arguably should be).

```scala
implicit class InfoMapSyntax[F[_]](info: Info[F, Map[Label.Name, String]]) {
  def info(labels: (Label.Name, String)*): F[Unit] = info.info(labels.toMap)
}
```

The extension class survives unchanged from v5. It's reachable now only when a consumer carries a Map-shaped `Info[F, Map[Label.Name, String]]` around — but in v6, to even get such a value, a consumer has to:

1. Declare labels at build time via `.label[String]("k1").label[String]("k2")...` — the new DSL forces this regardless.
2. `.contramap` the resulting typed `Info[F, (String, String)]` back into an `Info[F, Map[Label.Name, String]]`.
3. Then write `i.info(Label.Name("k1") -> "v1", Label.Name("k2") -> "v2")`.

Versus the idiomatic v6 migration:

1. Declare labels at build time (same step).
2. Observe with a tuple: `i.info(("v1", "v2"))`.

The `InfoMapSyntax` path is **strictly more work** — extra `.contramap`, extra `Label.Name(...)` wrapping at every call site, no type-safety win. There's no realistic migration scenario where a consumer benefits from keeping it.

Deleting it:
- Forces consumers onto the typed shape, which is the whole point of the redesign.
- Removes a piece of v5 vocabulary that future maintainers would otherwise have to explain ("why does this exist when the DSL doesn't produce it?").
- Closes a small migration confusion ("is `.info(Map(...))` still the right thing to do?").

_Theoretical exception_: a consumer with truly dynamic label key sets where keys aren't known at compile time. That pattern isn't supported by the v6 DSL anyway — such a consumer would be calling `MetricRegistry` directly and wouldn't be reaching for this extension class regardless.

PR #131, opened 2026-05-19
Add NHCB-friendly dual-mode histogram via .withNative

hgdyeo:
Adds NHCB-friendly dual-mode emission to the `.histogram(...)` builder via a new `.withNative` step. The promoted histogram emits **both** classic and native exponential representations from a single declaration.

This is the recommended migration target for existing histograms with curated bucket boundaries: classic representation preserves bucket intent for any Prometheus that doesn't speak native, AND Prometheus 2.49+'s `convert_classic_histograms_to_nhcb` performs server-side NHCB conversion from the classic form.

## API shape

```scala
// existing: classic-only (unchanged)
factory.histogram("http_dur").ofDouble.help("...")
  .buckets(0.005, 0.01, ..., 10).build

// new: dual-mode via .withNative (NHCB-friendly migration target)
factory.histogram("http_dur").ofDouble.help("...")
  .buckets(0.005, 0.01, ..., 10).withNative.build

// new: dual-mode with custom native tuning
factory.histogram("http_dur").ofDouble.help("...")
  .buckets(0.005, 0.01, ..., 10).withNative(NativeHistogram.Default.withInitialSchema(6)).build
```

Three histogram modes now available:

| Entrypoint | Classic emission | Native emission |
|---|---|---|
| `factory.histogram(...).buckets(...)` | ✓ (default) | ✗ |
| `factory.histogram(...).buckets(...).withNative` | ✓ (NHCB-friendly dual mode) | ✓ |
| `factory.nativeHistogram(...)` | ✗ | ✓ (exponential-only) |

## Summary

- `MetricRegistry.createAndRegisterDouble/LongHistogramWithNative` — new abstract methods (Long contramap-derived in `DoubleMetricRegistry`).
- New `HistogramWithNativeOps` mixin trait exposes `.withNative` on a histogram DSL.
- Two new concrete classes — `HistogramMetricDsl.Plain` (extends `MetricDsl`) and `HistogramMetricDsl.WithCallbacksImpl` (extends `MetricDsl.WithCallbacks`) — mix in the trait alongside the appropriate `MetricDsl` flavour.
- Compound type alias `HistogramMetricDsl[F, A] = MetricDsl[F, A, Histogram] with HistogramWithNativeOps[F, A]` exposes both DSL surfaces on a single value.
- `javasimpleclient` adapter throws `UnsupportedOperationException` (simpleclient can't emit native).
- The `javaclient` adapter dual-mode implementation lands in PR 7 alongside the classic + native-only implementations.

## Context

PR 5 of the v6 backend uplift split out of #124. Stacked on #130 (Info redesign).

## Why this PR is split out

The dual-mode design decision (`.withNative` as a separate step) is conceptually distinct from the native-only entrypoint in PR 3. Splitting them lets reviewers evaluate the API shape choice (`.withNative` after `.buckets(...)`, callback drop on promotion) in isolation.

## Risk

Additive — existing `.histogram(...).buckets(...).build` chains are bytecode-identical pre- and post-PR. Source-incompatible for external `MetricRegistry` implementers (must satisfy the two new abstract methods).

## Test plan

- [x] `sbt fix --check` clean

## Reviewer focus

- **`.withNative` promotion drops callback support.** Documented in scaladoc (the classic-only `Histogram.Value` snapshot type can't capture native bucket state). A consumer doing `.histogram(...).buckets(...).withNative.callback(...)` will get a compile error that isn't obviously about the callback drop. Worth a more prominent note on `MetricFactory.WithCallbacks.histogram`.
- **Long variant exists** (`.ofLong.buckets(...).withNative` works), but PR 3's native-only `.nativeHistogram(...)` has no Long variant. Inconsistency — should be reconciled.
- **`MetricsFactoryDslTest` coverage is thin.** Three dual-mode cases tested. Missing: the Long path (`.ofLong.buckets(...).withNative.build`), a regression test for `.callback(...)` on the non-promoted path, and a negative compile test (`assertDoesNotCompile`) for `.withNative.callback(...)`. Worth adding before this lands non-draft.
- **Wire-format implication** (for migration guide PR 11): dual-mode emits both classic and native bucket series per scrape — scrape payload grows. Consumers should be aware that the `convert_classic_histograms_to_nhcb` server-side option is the efficient way to consume this.

### DIFF
--- a/modules/prometheus4cats/src/main/scala/prometheus4cats/MetricFactory.scala
+++ b/modules/prometheus4cats/src/main/scala/prometheus4cats/MetricFactory.scala
@@ -558,11 +558,21 @@ object MetricFactory {
         )
       )
 
+    /** Type alias for the histogram DSL returned in the callback-aware variant: same shape as
+      * [[MetricFactory.HistogramWithNativeDsl]] but with the more specific [[HistogramMetricDsl.WithCallbacksImpl]] in
+      * place of the compound type, so consumers see `.callback(...)` as well as `.label(...)` / `.build` /
+      * `.withNative` after `.buckets(...)`.
+      *
+      * This is a subtype of `MetricFactory.HistogramWithNativeDsl[A]` (BucketDsl + HelpStep + TypeStep are all
+      * covariant in the right places), so the override is valid.
+      */
+    type HistogramWithCallbacksAndNativeDsl[A] = HelpStep[BucketDsl[HistogramMetricDsl.WithCallbacksImpl[F, A], A]]
+
     /** @inheritdoc */
-    override def histogram(name: Histogram.Name): TypeStep[HistogramWithNativeDsl] =
-      new TypeStep[HistogramWithNativeDsl](
+    override def histogram(name: Histogram.Name): TypeStep[HistogramWithCallbacksAndNativeDsl] =
+      new TypeStep[HistogramWithCallbacksAndNativeDsl](
         new HelpStep(help =>
-          new BucketDsl[HistogramMetricDsl[F, Long], Long](buckets =>
+          new BucketDsl[HistogramMetricDsl.WithCallbacksImpl[F, Long], Long](buckets =>
             new HistogramMetricDsl.WithCallbacksImpl[F, Long](
               classicMakeMetric = new LabelledMetricPartiallyApplied[F, Long, Histogram] {
 
@@ -600,7 +610,7 @@ object MetricFactory {
           )
         ),
         new HelpStep(help =>
-          new BucketDsl[HistogramMetricDsl[F, Double], Double](buckets =>
+          new BucketDsl[HistogramMetricDsl.WithCallbacksImpl[F, Double], Double](buckets =>
             new HistogramMetricDsl.WithCallbacksImpl[F, Double](
               classicMakeMetric = new LabelledMetricPartiallyApplied[F, Double, Histogram] {
 
--- a/modules/prometheus4cats/src/main/scala/prometheus4cats/MetricFactory.scala
+++ b/modules/prometheus4cats/src/main/scala/prometheus4cats/MetricFactory.scala
@@ -152,22 +152,47 @@ sealed abstract class MetricFactory[F[_]](
 
   type HistogramDsl[MDsl[_[_], _, _[_[_], _, _]], A] = HelpStep[BucketDsl[MDsl[F, A, Histogram], A]]
 
+  /** Type alias for the histogram DSL chain returned by [[histogram]]. After `.buckets(...)` the chain produces a value
+    * extending both `MetricDsl[F, A, Histogram]` and the `HistogramMetricDsl` mixin (which adds `.withNative` for
+    * dual-mode promotion). Subclasses (e.g., the WithCallbacks override) can swap in their own concrete impl that
+    * preserves the [[HistogramMetricDsl]] mixin.
+    */
+  type HistogramWithNativeDsl[A] = HelpStep[BucketDsl[HistogramMetricDsl[F, A], A]]
+
   /** Starts creating a "histogram" metric.
     *
+    * After declaring buckets, the resulting builder accepts an optional `.withNative` step that promotes the histogram
+    * from classic-only (default) to dual-mode emission (classic + native exponential — the NHCB-friendly configuration
+    * that preserves curated bucket boundaries while also enabling native histogram benefits).
+    *
     * @example
-    *   {{{ metrics.histogram("my_histogram").ofDouble.help("my counter help").buckets(1.0, 2.0)
-    *   .label[Int]("first_label").label[String]("second_label").label[Boolean]("third_label") .build }}}
+    *   {{{
+    * // Classic-only (existing behaviour):
+    * metrics.histogram("http_dur").ofDouble.help("...").buckets(0.005, 0.01, 0.025, 0.05, 0.1, 0.5, 1, 5, 10).build
+    *
+    * // Dual-mode (NHCB-friendly): same buckets, plus native exponential.
+    * metrics.histogram("http_dur").ofDouble.help("...")
+    *   .buckets(0.005, 0.01, 0.025, 0.05, 0.1, 0.5, 1, 5, 10)
+    *   .withNative
+    *   .build
+    *
+    * // Dual-mode with custom native tuning.
+    * metrics.histogram("http_dur").ofDouble.help("...")
+    *   .buckets(0.005, 0.01, ...)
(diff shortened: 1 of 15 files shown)


modules/prometheus4cats/src/test/scala/test/MetricsFactoryDslTest.scala
```diff
@@ -194,6 +194,18 @@ class MetricsFactoryDslTest[F[_]: MonadCancelThrow: Clock] {
...
+
+  histogramBuilder.ofDouble.help("dual").buckets(0.1, 0.5, 1.0).withNative.label[String]("label1").build
+
+  histogramBuilder.ofDouble
+    .help("dual tuned")
+    .buckets(0.1, 0.5, 1.0)
+    .withNative(NativeHistogram.Default.withInitialSchema(6))
+    .label[String]("label1")
+    .build
+
```
hgdyeo, line 208: **Optional / non-blocking** — three small additions worth considering for this file, all compile-time only:

1. **Long path coverage.** The three new dual-mode cases are all `.ofDouble`. The `.ofLong.buckets(...).withNative` path goes through the `DoubleMetricRegistry` contramap derivation and isn't currently exercised:
   ```scala
   histogramBuilder.ofLong.help("dual long").buckets(1L, 10L, 100L).withNative.build
   ```

2. **Callback regression on the non-promoted path.** Commit [7e3d790](https://github.com/permutive-engineering/prometheus4cats/pull/131/commits/7e3d790f1941b5086aa4e624cf5a7140f07071f5)'s `HistogramWithCallbacksAndNativeDsl` plumbing fix keeps `.callback` reachable on `.histogram(...).buckets(...).callback(...).build` in the `WithCallbacks` factory. No test currently asserts that fix is still in place — a future refactor could erase it silently:
   ```scala
   factoryWithCallbacks.histogram("h").ofDouble.help("...")
     .buckets(0.1, 0.5, 1.0)
     .callback(IO.pure(NonEmptyList.of(Histogram.Value(...) -> "label-value")))
     .label[String]("label1")
     .build
   ```

3. **Compile-time rejection of `.withNative.callback`.** The deliberate API restriction (callback dropped on dual-mode promotion) isn't currently asserted as a compile error. A negative test would catch accidental re-enablement:
   ```scala
   // assertDoesNotCompile / compileErrors equivalent:
   factoryWithCallbacks.histogram("h").ofDouble.help("...")
     .buckets(0.1, 0.5, 1.0).withNative.callback(???).build
   ```

None of these block merge — entirely happy to leave it as-is and treat the DSL invariants as code-review-only. Just flagging because the plumbing fix is subtle and the deliberate callback gap is the kind of thing future refactors lose track of.

hgdyeo (commented):
One optional / non-blocking suggestion on DSL test coverage. Inline below.

kidonm (commented):
Code review with 2 comments, both non-blocking.

modules/prometheus4cats/src/main/scala/prometheus4cats/MetricFactory.scala
```diff
@@ -511,16 +558,23 @@ object MetricFactory {
...
+      *
+      * This is a subtype of `MetricFactory.HistogramWithNativeDsl[A]` (BucketDsl + HelpStep + TypeStep are all
+      * covariant in the right places), so the override is valid.
+      */
+    type HistogramWithCallbacksAndNativeDsl[A] = HelpStep[BucketDsl[HistogramMetricDsl.WithCallbacksImpl[F, A], A]]
 
     /** @inheritdoc */
-    override def histogram(name: Histogram.Name): TypeStep[HistogramDsl[HistogramCallbackDsl, *]] =
-      new TypeStep[HistogramDsl[HistogramCallbackDsl, *]](
+    override def histogram(name: Histogram.Name): TypeStep[HistogramWithCallbacksAndNativeDsl] =
```
kidonm, line 572: **thought** (non-blocking): `.withNative` silently drops callback support — worth a chat on how visible we want that to be, and whether to revisit later.

Once a consumer goes through `.buckets(...).withNative`, the resulting DSL is a plain `MetricDsl` (no `.callback(...)`), because the existing callback machinery is typed over `Histogram.Value[A]` — a classic-only snapshot type (sum, count, bucket counts). To support `.withNative.callback(...)` you would need a richer snapshot type carrying schema / zero-threshold / sparse positive+negative buckets, new `CallbackRegistry` SPI methods, and matching backend implementations. So this is a library type-system scope decision, not a wire-format or upstream limitation — `prometheus-metrics-core` 1.x can emit native histograms from a pulled snapshot just fine.

Two things that might soften the rough edge in the meantime:

1. A more prominent scaladoc note on `MetricFactory.WithCallbacks.histogram` (and on `withNative` itself) calling out that promotion drops callback support, so consumers do not have to reverse-engineer the compile error.
2. An `assertDoesNotCompile` test pinning the behaviour, so a future refactor that accidentally restores `.callback` post-`.withNative` does not slip through.

Separately, worth a quick conversation about whether pull-based native histograms are useful enough to be on the long-term roadmap — they fit classic histograms more naturally (snapshot a few sums on demand) than native (which shines on cheap high-resolution accumulation).

modules/prometheus4cats/src/main/scala/prometheus4cats/MetricRegistry.scala
```diff
@@ -205,6 +205,38 @@ trait MetricRegistry[F[_]] {
...
+    * [[cats.effect.kernel.Resource]] (the simpleclient backend cannot emit native histograms).
+    */
+  def createAndRegisterDoubleHistogramWithNative[A](
+      prefix: Option[Metric.Prefix],
+      name: Histogram.Name,
+      help: Metric.Help,
+      commonLabels: Metric.CommonLabels,
+      labelNames: IndexedSeq[Label.Name],
+      buckets: NonEmptySeq[Double],
+      nativeHistogram: NativeHistogram
```
kidonm, line 226: **chore** (non-blocking): rename `nativeHistogram` → `config` here (and on `createAndRegisterLongHistogramWithNative`, plus the simpleclient/testing/mapK/external-test implementations) for consistency with #129's recent rename.

On the parent branch, commit b79f6b0 ("Address PR #129 review: rename nativeHistogram→config") renamed the same parameter on the native-only SPI to `config`. This PR re-introduces the old name on the dual-mode SPI, so the two methods now read:

```scala
def createAndRegisterDoubleNativeHistogram[A](..., config: NativeHistogram)              // #129
def createAndRegisterDoubleHistogramWithNative[A](..., nativeHistogram: NativeHistogram) // #131
```

Easiest to fix while #131 is still rebased on #129 — once both land the inconsistency becomes a separate cleanup.

PR #132, opened 2026-05-20
Add javaclient backend skeleton with Counter and Gauge

hgdyeo:
First piece of the v6 backend uplift: introduces the `prometheus4cats.javaclient` package as a sibling of the legacy `javasimpleclient`. Counter and Gauge are fully implemented against `prometheus-metrics-core` 1.x; every other registration method raises `UnsupportedOperationException("not yet ported")` and will be filled in by subsequent PRs.

## Summary

New package `modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/`:

- `package.scala` — state-machine type aliases over upstream's `StatefulMetric`.
- `models/MetricType.scala`, `models/Exceptions.scala`.
- `internal/Utils.scala` — `modifyMetric` + label-values DataPoint helper.
- `JavaMetricRegistry.scala` — Counter (Double) + Gauge (Double) fully implemented against the 1.x `builder()` / `labelValues()` API. Long variants come for free via the `DoubleMetricRegistry` mixin. All other abstract methods (Histogram, Summary, Info, callbacks, MetricCollection) raise `UnsupportedOperationException("not yet ported")`.
- `JavaMetricRegistry.Builder` — mirrors the legacy `javasimpleclient` Builder API so consumers can migrate by changing only the import. Differences: takes a `PrometheusRegistry` instead of `CollectorRegistry`; `withJvmMetrics` replaces `withHotSpotCollectors` (uses upstream's one-line `JvmMetrics.builder().register(...)`).

The `configureBuilderOrRetrieve` helper preserves the legacy adapter's reference-counted registration semantics — overlapping `Resource`-scoped registrations of the same metric name reuse the same underlying collector and only unregister on the last release.

The new package coexists with `javasimpleclient` — disjoint package namespaces (`io.prometheus.metrics.*` vs `io.prometheus.client.*`) mean no classpath conflict. Both backends compile and pass their own tests in this PR.

## Context

PR 6 of the v6 backend uplift split out of #124. Stacked on #131 (`.withNative` dual-mode).

Includes a small follow-up commit (`Align javaclient skeleton stubs with v6 SPI`) that:

- Adds a throwing stub for `createAndRegisterDoubleHistogramWithNative` (introduced in PR 5).
- Updates `createAndRegisterInfo` to the v6 typed-labels signature (introduced in PR 4).
- Fully qualifies the `[[PrometheusRegistry]]` scaladoc links to `[[io.prometheus.metrics.model.registry.PrometheusRegistry]]` (matching the pattern the legacy adapter uses for `[[io.prometheus.client.CollectorRegistry]]`). Unqualified `[[link]]` references to imported Java types don't resolve in scaladoc.

These adjustments are necessary because the original commit predates PR 4 + PR 5 in #124's history.

## Why this PR is split out

Counter and Gauge are mechanical wrappings of upstream's `builder()` chain — the real review surface is `configureBuilderOrRetrieve` (ref-counted registration) and the `Builder` API parity with the legacy adapter. Isolating these in one PR lets reviewers focus on the skeleton's invariants before the larger Histogram (PR 7) and Callback (PR 9) work piles on.

## Risk

Low — package coexists with legacy. The legacy adapter and its test suite still compile and pass unchanged. All non-Counter / non-Gauge methods are stubs that throw `UnsupportedOperationException`, so no consumer can accidentally use them yet.

## Test plan

- [x] `sbt ci-test` green locally on Scala 2.13.18 + 3.3.7 (114 existing tests pass; no new tests in this PR — smoke tests come with the Histogram impl in PR 7).

## Reviewer focus

- **`configureBuilderOrRetrieve`** — preserves the legacy adapter's ref-counted registration semantics. Overlapping `Resource`-scoped registrations of the same metric name reuse the same underlying collector and only unregister on last release.
- **Builder API parity** with the legacy `Builder` for drop-in replacement at import sites. `.withRegistry` now takes `PrometheusRegistry`; `withJvmMetrics` replaces `withHotSpotCollectors`.
- **The "not yet ported" stubs.** These will be replaced one-by-one in PRs 7-9. Choosing `UnsupportedOperationException` over silent no-op so any accidental call surfaces loudly.

### DIFF
--- a/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
+++ b/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
@@ -281,6 +281,70 @@ class JavaMetricRegistry[F[_]: Async] private (
     }
   }
 
+  override def createAndRegisterDoubleHistogramWithNative[A](
+      prefix: Option[Metric.Prefix],
+      name: Histogram.Name,
+      help: Metric.Help,
+      commonLabels: Metric.CommonLabels,
+      labelNames: IndexedSeq[Label.Name],
+      buckets: NonEmptySeq[Double],
+      nativeHistogram: NativeHistogram
+  )(f: A => IndexedSeq[String]): Resource[F, Histogram[F, Double, A]] = {
+    val commonLabelNames       = commonLabels.value.keys.toIndexedSeq
+    val commonLabelValuesArray = commonLabels.value.values.toArray
+    val allLabelNames          = labelNames ++ commonLabelNames
+    val fullName               = NameUtils.makeName(prefix, name)
+
+    configureBuilderOrRetrieve[PHistogram](
+      register = () => {
+        // Dual-mode: NEITHER .classicOnly() NOR .nativeOnly(). Both classicUpperBounds(...) and
+        // nativeInitialSchema(...) are set. The resulting Histogram emits BOTH representations,
+        // letting Prometheus's `convert_classic_histograms_to_nhcb` pick the classic form for
+        // server-side NHCB conversion while the native exponential is also available directly.
+        val builder = PHistogram
+          .builder()
+          .name(fullName)
+          .help(help.value)
+          .labelNames(allLabelNames.map(_.value): _*)
+          .classicUpperBounds(buckets.toSeq.toArray: _*)
+          .nativeInitialSchema(nativeHistogram.initialSchema)
+          .nativeMaxNumberOfBuckets(nativeHistogram.maxNumberOfBuckets)
+          .nativeMaxZeroThreshold(nativeHistogram.maxZeroThreshold)
+          .nativeMinZeroThreshold(nativeHistogram.minZeroThreshold)
+        val tuned =
+          if (nativeHistogram.resetDuration > 0.seconds)
+            builder.nativeResetDuration(
+              nativeHistogram.resetDuration.toSeconds,
+              java.util.concurrent.TimeUnit.SECONDS
+            )
+          else builder
+        tuned.register(registry)
+      },
+      // Use a distinct MetricType so dedup is correct: registering the same metric name as
+      // dual-mode and then again as classic-only is a programmer error and should fail.
+      metricType = MetricType.HistogramWithNative,
+      metricPrefix = prefix,
+      stringName = name.value,
+      labels = allLabelNames
+    ).map { case (histogram, exemplarRef) =>
+      Histogram.make[F, Double, A](
+        Histogram.ExemplarState.fromRef(buckets, exemplarRef),
+        _observe = { (d: Double, labels: A, exemplar: Option[Exemplar.Labels]) =>
+          Utils.modifyMetric[F, Histogram.Name, io.prometheus.metrics.core.datapoints.DistributionDataPoint](
+            metricName = name,
+            allLabelNames = allLabelNames,
+            dynamicLabels = f(labels),
+            commonLabelValues = commonLabelValuesArray,
+            getDataPoint = (lbls: Array[String]) => histogram.labelValues(lbls: _*),
+            modify = (dp: io.prometheus.metrics.core.datapoints.DistributionDataPoint) =>
+              exemplar.fold(dp.observe(d))(e => dp.observeWithExemplar(d, transformExemplarLabels(e))),
+            logger = logger
+          )
+        }
+      )
+    }
+  }
+
   override def createAndRegisterDoubleNativeHistogram[A](
       prefix: Option[Metric.Prefix],
       name: Histogram.Name,
@@ -342,17 +406,6 @@ class JavaMetricRegistry[F[_]: Async] private (
       )
     }
   }
-
-  override def createAndRegisterDoubleHistogramWithNative[A](
-      prefix: Option[Metric.Prefix],
-      name: Histogram.Name,
-      help: Metric.Help,
-      commonLabels: Metric.CommonLabels,
-      labelNames: IndexedSeq[Label.Name],
-      buckets: NonEmptySeq[Double],
-      nativeHistogram: NativeHistogram
-  )(f: A => IndexedSeq[String]): Resource[F, Histogram[F, Double, A]] =
-    Resource.eval(ApplicativeThrow[F].raiseError(notYetPorted("createAndRegisterDoubleHistogramWithNative")))
 
   override def createAndRegisterDoubleSummary[A](
       prefix: Option[Metric.Prefix],
--- a/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
+++ b/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
@@ -153,7 +153,6 @@ class JavaMetricRegistry[F[_]: Async] private (
           .name(fullName)
           .help(help.value)
           .labelNames(allLabelNames.map(_.value): _*)
-          .withExemplars()
           .register(registry),
       metricType = MetricType.Counter,
       metricPrefix = prefix,
--- a/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
+++ b/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
@@ -45,6 +45,7 @@ import io.prometheus.metrics.core.metrics.{Summary => PSummary}
 import io.prometheus.metrics.instrumentation.jvm.JvmMetrics
 import io.prometheus.metrics.model.registry.PrometheusRegistry
 import io.prometheus.metrics.model.registry.{Collector => PCollector}
+import io.prometheus.metrics.model.registry.{MultiCollector => PMultiCollector}
 import io.prometheus.metrics.model.snapshots.ClassicHistogramBuckets
 import io.prometheus.metrics.model.snapshots.CounterSnapshot
 import io.prometheus.metrics.model.snapshots.DataPointSnapshot
@@ -54,6 +55,7 @@ import io.prometheus.metrics.model.snapshots.HistogramSnapshot
 import io.prometheus.metrics.model.snapshots.Labels
 import io.prometheus.metrics.model.snapshots.MetricMetadata
 import io.prometheus.metrics.model.snapshots.MetricSnapshot
+import io.prometheus.metrics.model.snapshots.MetricSnapshots
 import io.prometheus.metrics.model.snapshots.Quantiles
 import io.prometheus.metrics.model.snapshots.SummarySnapshot
 import io.prometheus.metrics.model.snapshots.{Quantile => PQuantile}
@@ -83,6 +85,20 @@ class JavaMetricRegistry[F[_]: Async] private (
     private val dispatcher: Dispatcher[F],
     private val callbackTimeout: FiniteDuration,
     private val singleCallbackTimeout: FiniteDuration,
+    // Metric-collection callbacks: keyed by prefix (each prefix has its own merged common-labels map
+    // and a token-keyed map of user callbacks producing MetricCollection values). Constructed
+    // here, but the upstream MultiCollector that aggregates them is built in the class body via
+    // [[buildMetricCollectionMultiCollector]] so it can capture `this`.
+    private val metricCollectionRef: Ref[F, Map[
+      Option[Metric.Prefix],
+      (
+          Map[Label.Name, String],
+          Map[
+            Unique.Token,
+            F[MetricCollection]
+          ]
+      )
+    ]],
     private val logger: Throwable => String => F[Unit]
 ) extends DoubleMetricRegistry[F]
     with DoubleCallbackRegistry[F] {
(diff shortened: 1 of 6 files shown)


hgdyeo (commented):
Four review comments — two blockers (forward-portable fixes from PR 7 of #124), one Major (a catch-all lint suppression that's never removed later in #124), one Nit. Detail inline.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -0,0 +1,428 @@
...
+    val n                      = counterName(name)
+    val fullName               = NameUtils.makeName(prefix, name)
+
+    configureBuilderOrRetrieve[PCounter](
+      register = () =>
+        PCounter
+          .builder()
+          .name(fullName)
+          .help(help.value)
+          .labelNames(allLabelNames.map(_.value): _*)
```
hgdyeo, line 212: **[Blocker] `.withExemplars()` is doubly redundant — broken AND a no-op even if it worked.** Recommend dropping the call (forward-port from PR 7 of #124, commit `edebc8d`).

Verified against upstream [`StatefulMetric.java` on main](https://github.com/prometheus/client_java/blob/main/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/StatefulMetric.java):

**1. The call is broken (IllegalAccessError).** `withExemplars()` is declared on `StatefulMetric.Builder`, which is package-private (no access modifier):

```java
abstract static class Builder<B extends Builder<B, M>, M extends StatefulMetric<?, ?>>
    extends MetricWithFixedMetadata.Builder<B, M> {
  ...
  public B withExemplars() { ... }
}
```

After erasure, scalac emits bytecode referencing `StatefulMetric$Builder` as the static return type. From outside `io.prometheus.metrics.core.metrics`, the JVM access check on that return-type class fails with `IllegalAccessError` the first time a Counter is registered.

**2. Even if it worked, it would do nothing.** Exemplars default to enabled in 1.x:

```java
/** Override if there are more properties than just exemplars enabled. */
public MetricsProperties getDefaultProperties() {
  return MetricsProperties.builder().exemplarsEnabled(true).build();
}
```

**Suggested change:** remove line 155. One-line fix, no behavioural change to the constructed `PCounter` — exemplars stay enabled via the default.

The bug isn't exercised in PR 6 (no Counter tests until PR 7) but any consumer using the new backend between PR 6 merge and PR 7 merge will hit `IllegalAccessError` on first Counter registration. Recommend forward-porting now.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -0,0 +1,428 @@
...
+          .withExemplars()
+          .register(registry),
+      metricType = MetricType.Counter,
+      metricPrefix = prefix,
+      stringName = n,
+      labels = allLabelNames
+    ).flatMap { counter =>
+      Resource.eval(Ref.of[F, Option[Exemplar.Data]](None)).map { exemplarRef =>
+        Counter.make(
+          Counter.ExemplarState.fromRef(exemplarRef),
```
hgdyeo: **[Blocker] Exemplar `Ref` divergence** — two `Resource`-scoped acquisitions of the same Counter end up with two unrelated exemplar trackers, even though they correctly share one underlying `PCounter`. Recommend forward-porting the fix from PR 7 of #124 (commit `edebc8d`).

## The bug, step by step

**Step 1 — `configureBuilderOrRetrieve` creates a tracker and stores it in state** (lines 111-115):

```scala
case None =>
  for {
    exemplarRef <- Ref.of[F, Option[Exemplar.Data]](None)   // (A) create
    collector   <- Sync[F].delay(register())
  } yield (metrics.updated(fullName, (metricId, (collector, exemplarRef, 1))), collector)
//                                                          ^^^^^^^^^^^^^
//                                                          stored in state
```

**Step 2 — but it returns only the collector** (line 88):

```scala
): Resource[F, M] = {
//   ^^^^^^^^^^^^
//   only M, not (M, exemplarRef)
```

The tracker `(A)` is stored in state but never returned to the caller.

**Step 3 — `createAndRegisterDoubleCounter` creates a SECOND tracker** (lines 162-167, this comment's anchor):

```scala
).flatMap { counter =>
  Resource.eval(Ref.of[F, Option[Exemplar.Data]](None)).map { exemplarRef =>   // (B) another tracker
    Counter.make(
      Counter.ExemplarState.fromRef(exemplarRef),                              // bound to (B)
```

A fresh `Ref`, completely unrelated to `(A)`. The Counter wrapper's sampler is bound to `(B)`. `(A)` in state is orphaned.

## Consequence

```scala
val makeCounter = factory.counter("requests_total").ofDouble.help("...").label[String]("path").build

makeCounter.use { counterA =>
  makeCounter.use { counterB =>
    // counterA and counterB wrap the SAME underlying PCounter (deduplicated correctly).
    // But each has its OWN exemplar tracker. Exemplars sampled via view A aren't
    // visible to view B and vice versa — divergent state.
  }
}
```

The legacy `javasimpleclient` adapter stores the exemplar Ref in state and returns it so the SAME ref is reused across duplicate acquisitions. PR 6 silently diverges.

## Fix

From PR 7's commit body (`edebc8d`):

> Subtle fix to `configureBuilderOrRetrieve`: now returns `Resource[F, (M, Ref[F, Option[Exemplar.Data]])]` rather than `Resource[F, M]`. The exemplar Ref is stored in State and shared across duplicate registrations of the same metric, matching legacy `javasimpleclient` behaviour.

Four changes:

1. Change return type at line 88 to `Resource[F, (M, Ref[F, Option[Exemplar.Data]])]`.
2. Update the existing-entry branch (line 96-103) to return the stored `exemplarRef` alongside the collector.
3. Update the new-entry branch (line 111-115) to return the freshly-created `exemplarRef` alongside the collector.
4. In `createAndRegisterDoubleCounter` (line 162-167), use the returned `exemplarRef` instead of creating a new one — drop the `Resource.eval(Ref.of...)` step entirely.

~10 lines of changes. Recommend forward-porting to keep `series/6.x` correct between PR 6 and PR 7 merges.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -0,0 +1,428 @@
...
+
+/** Implements [[MetricRegistry]] and [[CallbackRegistry]] against the upstream `prometheus-metrics-core` 1.x library
+  * (the successor to the legacy simpleclient backend).
+  *
+  * This class is the v6 replacement for `prometheus4cats.javasimpleclient.JavaMetricRegistry`. Both implementations
+  * coexist during the migration window.
+  *
+  * Construct via [[JavaMetricRegistry.Builder]].
+  */
+@SuppressWarnings(Array("all"))
```
hgdyeo: **[Major] `@SuppressWarnings(Array("all"))` is suspiciously broad — and never removed later in #124's history.**

`Array("all")` disables every scalafix / wartremover / linter warning on this class (and the companion object at line 344). Verified: the annotation is **introduced here in the skeleton commit and is still present at #124's final tip** (`fee16ad`). So this isn't deferred cleanup — it's a deliberate decision to make now.

What the catch-all is likely hiding:

- `asInstanceOf[M]` cast at line 102 (Wart `AsInstanceOf`).
- Raw `RuntimeException` via `raiseError` (Wart `Throw`).
- `getClass` usage in `Utils.scala` (which already has its own `// scalafix:ok` workaround — see separate comment).

The legacy `javasimpleclient/JavaMetricRegistry.scala:62` does the same thing, so this is mirroring an existing pattern — but landing v6 with the same lint-silencing as v5 perpetuates the smell into the next major.

**Suggested change:** remove the annotation, run `sbt fix`, see what fires, re-add targeted per-warning suppressions (e.g. `@SuppressWarnings(Array("scalafix:DisableSyntax.asInstanceOf"))`) for things that genuinely can't be fixed.

**Push back if:** matching the legacy adapter's pattern is more important than cleaning up. In that case, worth opening a follow-up issue to clean up both adapters before v6 ships.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/internal/Utils.scala
```diff
@@ -0,0 +1,94 @@
...
+  ): F[B] =
+    fa.handleErrorWith(e =>
+      classStringRep(e)
+        .flatMap(className =>
+          Sync[F].raiseError(
+            UnhandledPrometheusException(className, metricName, labelNames.zip(labels.toIndexedSeq).toMap, e)
+          )
+        )
+    )
+
```
hgdyeo, line 90: **[Nit, not blocking]** `a.getClass.toString` is pure for non-null `a`, and the only call site (`handleErrors`) passes a `Throwable` from `handleErrorWith` (guaranteed non-null). So the `Sync.delay` and the `// scalafix:ok` aren't earning their keep here — the helper could be inlined:

```scala
fa.handleErrorWith(e =>
  Sync[F].raiseError(
    UnhandledPrometheusException(e.getClass.toString, metricName, ..., e)
  )
)
```

Drops the helper and the `scalafix:ok` workaround. Happy to keep as-is if you prefer the structural symmetry of "all operations go through `F`".

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
hgdyeo: Addressed in 0e15fea — dropped the `.withExemplars()` call. Confirmed via `sbt ci-test` locally on Scala 2.13 + 3.3.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
hgdyeo: Addressed in b1dfa2a — `configureBuilderOrRetrieve` now returns `Resource[F, (M, Ref[F, Option[Exemplar.Data]])]` so the exemplar Ref stored in state propagates to the caller. `createAndRegisterDoubleCounter` destructures the pair and uses the propagated Ref; the redundant `Resource.eval(Ref.of...)` step is dropped. Gauge updated to destructure too. All 114 tests pass.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
hgdyeo: Addressed in cdcac3e — removed both blanket `@SuppressWarnings(Array("all"))` annotations from the class and companion. Running `sbt fix` surfaced exactly one violation: `scalafix:DisableSyntax.==` on the `metricId == expected` comparison in `configureBuilderOrRetrieve` (both sides are `(IndexedSeq[Label.Name], MetricType)` — known same type, `Tuple2.equals` is correct). Replaced with a targeted `@SuppressWarnings(Array("scalafix:DisableSyntax.=="))` on that one method only.

Note: the legacy `javasimpleclient/JavaMetricRegistry.scala:61` has the same blanket pattern. Left untouched as out-of-scope for this PR.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/internal/Utils.scala
hgdyeo: Not addressing — the same `Sync.delay(... .getClass.toString) // scalafix:ok` pattern is used in the equivalent helper of the legacy `javasimpleclient` adapter (`javasimpleclient/internal/Utils.scala`). Cleaning it up in the new adapter while leaving the legacy adapter as-is would create an asymmetry mid-migration; a separate follow-up could clean up both adapters together. Leaving as-is for this PR.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -0,0 +1,424 @@
...
+
+    /** Register the standard JVM/process metrics (memory pools, GC, threads, class loading, buffers) via the upstream
+      * `prometheus-metrics-instrumentation-jvm` library when the registry is built. Replacement for the legacy
+      * `withHotSpotCollectors` builder method.
+      */
+    def withJvmMetrics: Builder[F] = copy(registerJvmMetrics = true)
+
+    def build: Resource[F, JavaMetricRegistry[F]] =
+      Resource.eval {
+        if (registerJvmMetrics) Sync[F].delay(JvmMetrics.builder().register(promRegistry))
```
kidonm, line 1249: **issue** (non-blocking): JVM metrics registration outlives the `JavaMetricRegistry` `Resource`, which breaks repeated builds against the same `PrometheusRegistry`.

`JvmMetrics.builder().register(promRegistry)` runs inside a `Resource.eval` whose release is `unit`. The inner `Resource.make(acquire)(...)` cleanup only unregisters the Counter/Gauge collectors that this registry tracked — JVM metrics are not in that map and stay registered on `promRegistry` forever. Building a second `JavaMetricRegistry` against the same `PrometheusRegistry` (the default singleton is a common case, especially in tests) with `.withJvmMetrics` will then fail on the second call with upstream's "already registered" error.

Cleanest fix would be to scope the JVM-metrics registration as a proper `Resource.make` that unregisters on release, but I noticed upstream's `JvmMetrics` is a facade that registers many individual collectors and does not expose a clean "unregister all" — so the actual fix is probably one of:

1. Track the individual collectors returned by `JvmMetrics`' inner registrations and unregister each on release, or
2. Make the registration idempotent (no-op if JVM metrics are already on the target registry), or
3. Document the constraint loudly — "use a fresh `PrometheusRegistry` per `JavaMetricRegistry` build when `.withJvmMetrics` is set" — and leave the behaviour as-is.

Mostly want to flag so we go in eyes-open; happy with any of the three.

kidonm (commented):
Code review with 1 comment, non-blocking.

PR #133, opened 2026-05-21
Implement Histogram (classic, native, dual-mode) in javaclient

hgdyeo:
Replaces the three throwing stubs in the `javaclient` backend with real histogram implementations against `prometheus-metrics-core` 1.x. Covers all three modes from the SPI:

| Mode | Upstream builder | Where |
|---|---|---|
| Classic-only | `.classicOnly().classicUpperBounds(...)` | `createAndRegisterDoubleHistogram` |
| Native-only | `.nativeOnly().nativeInitialSchema(...)` etc. | `createAndRegisterDoubleNativeHistogram` |
| Dual-mode | neither (upstream default) | `createAndRegisterDoubleHistogramWithNative` |

## Summary

- **`JavaMetricRegistry.scala`** — three real impls replace the three `notYetPorted` stubs. All use the shared `configureBuilderOrRetrieve` helper (PR 132's exemplar Ref propagation) for ref-counted registration. Classic + dual use `Histogram.ExemplarState.fromRef(buckets, exemplarRef)` for bucket-driven exemplar sampling; native-only uses `ExemplarState.noop` (no bucket boundaries to sample against).
- **`models/MetricType.scala`** — adds `HistogramWithNative` enum case so the dedup key correctly distinguishes "registered as classic-only" from "registered as dual-mode" with the same metric name.
- **`JavaMetricRegistrySuite.scala`** — new test file with focused smoke tests: classic histogram with observation/snapshot assertion; native histogram with `hasNativeHistogramData` snapshot assertion; dual-mode asserting BOTH `hasClassicHistogramData` AND `hasNativeHistogramData` plus preserved bucket boundaries; tuning-override path via `NativeHistogram.Default.withInitialSchema(...)`.

## Context

PR 7 of the v6 backend uplift split out of #124. Stacked on #132 (javaclient skeleton). When PR 132 lands on `series/6.x`, this PR auto-rebases.

Two source commits combined:
- `edebc8d` — classic + native-only implementations + the test file.
- `425e542d`'s `javaclient/` hunks (originally carved out to PR 5 as core-only) — dual-mode implementation + the `HistogramWithNative` enum case + dual-mode smoke test.

## Why this PR is split out

Histograms are the single biggest implementation surface in the backend. Three modes × the upstream `Histogram.Builder` API + exemplar-Ref propagation gymnastics + the dedup-key extension — much easier to review as one focused PR than mixed with Counter/Gauge skeleton (PR 132) or Summary/Info (PR 8).

## Risk

Low. Histograms are a pure extension over PR 132's skeleton — three new methods on the same class. No changes to the SPI shape (committed in PRs 129/131). Test coverage proves the wire format end-to-end via `hasClassicHistogramData` / `hasNativeHistogramData` snapshot assertions.

## Test plan

- [x] `sbt ci-test` green locally on Scala 2.13.18 + 3.3.7 (121 tests pass, +7 new histogram smoke tests).

## Reviewer focus

- **`.classicOnly()` preserves v5 wire format.** Without this call, upstream's `Histogram.Builder` defaults to emitting BOTH representations — that would silently double the scrape payload for every existing histogram. Explicit `.classicOnly()` on the regular `.histogram(...)` path is the v5 compatibility guarantee.
- **`nativeResetDuration` is conditional.** Only added to the builder chain when `config.resetDuration > 0.seconds` — broken out into a separate val to avoid a discarded non-Unit return on the Scala 3 build.
- **Native exemplars use `ExemplarState.noop`.** The bucket-driven exemplar sampler doesn't apply when there are no bucket boundaries. Explicit `.observeWithExemplar(...)` still works end-to-end via the upstream `DistributionDataPoint` API. Sampler-based exemplars on native histograms can be added later if a consumer asks.
- **`MetricType.HistogramWithNative`** distinguishes dual-mode registrations from classic-only at the dedup-key level. Two registrations of the same name with different modes correctly raise the "different labels and/or type" error.

### DIFF
--- a/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
+++ b/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
@@ -281,6 +281,70 @@ class JavaMetricRegistry[F[_]: Async] private (
     }
   }
 
+  override def createAndRegisterDoubleHistogramWithNative[A](
+      prefix: Option[Metric.Prefix],
+      name: Histogram.Name,
+      help: Metric.Help,
+      commonLabels: Metric.CommonLabels,
+      labelNames: IndexedSeq[Label.Name],
+      buckets: NonEmptySeq[Double],
+      nativeHistogram: NativeHistogram
+  )(f: A => IndexedSeq[String]): Resource[F, Histogram[F, Double, A]] = {
+    val commonLabelNames       = commonLabels.value.keys.toIndexedSeq
+    val commonLabelValuesArray = commonLabels.value.values.toArray
+    val allLabelNames          = labelNames ++ commonLabelNames
+    val fullName               = NameUtils.makeName(prefix, name)
+
+    configureBuilderOrRetrieve[PHistogram](
+      register = () => {
+        // Dual-mode: NEITHER .classicOnly() NOR .nativeOnly(). Both classicUpperBounds(...) and
+        // nativeInitialSchema(...) are set. The resulting Histogram emits BOTH representations,
+        // letting Prometheus's `convert_classic_histograms_to_nhcb` pick the classic form for
+        // server-side NHCB conversion while the native exponential is also available directly.
+        val builder = PHistogram
+          .builder()
+          .name(fullName)
+          .help(help.value)
+          .labelNames(allLabelNames.map(_.value): _*)
+          .classicUpperBounds(buckets.toSeq.toArray: _*)
+          .nativeInitialSchema(nativeHistogram.initialSchema)
+          .nativeMaxNumberOfBuckets(nativeHistogram.maxNumberOfBuckets)
+          .nativeMaxZeroThreshold(nativeHistogram.maxZeroThreshold)
+          .nativeMinZeroThreshold(nativeHistogram.minZeroThreshold)
+        val tuned =
+          if (nativeHistogram.resetDuration > 0.seconds)
+            builder.nativeResetDuration(
+              nativeHistogram.resetDuration.toSeconds,
+              java.util.concurrent.TimeUnit.SECONDS
+            )
+          else builder
+        tuned.register(registry)
+      },
+      // Use a distinct MetricType so dedup is correct: registering the same metric name as
+      // dual-mode and then again as classic-only is a programmer error and should fail.
+      metricType = MetricType.HistogramWithNative,
+      metricPrefix = prefix,
+      stringName = name.value,
+      labels = allLabelNames
+    ).map { case (histogram, exemplarRef) =>
+      Histogram.make[F, Double, A](
+        Histogram.ExemplarState.fromRef(buckets, exemplarRef),
+        _observe = { (d: Double, labels: A, exemplar: Option[Exemplar.Labels]) =>
+          Utils.modifyMetric[F, Histogram.Name, io.prometheus.metrics.core.datapoints.DistributionDataPoint](
+            metricName = name,
+            allLabelNames = allLabelNames,
+            dynamicLabels = f(labels),
+            commonLabelValues = commonLabelValuesArray,
+            getDataPoint = (lbls: Array[String]) => histogram.labelValues(lbls: _*),
+            modify = (dp: io.prometheus.metrics.core.datapoints.DistributionDataPoint) =>
+              exemplar.fold(dp.observe(d))(e => dp.observeWithExemplar(d, transformExemplarLabels(e))),
+            logger = logger
+          )
+        }
+      )
+    }
+  }
+
   override def createAndRegisterDoubleNativeHistogram[A](
       prefix: Option[Metric.Prefix],
       name: Histogram.Name,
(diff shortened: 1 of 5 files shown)


hgdyeo (commented):
Three should-fix comments, all non-blocking. Inline below.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -233,29 +238,174 @@ class JavaMetricRegistry[F[_]: Async] private (
...
+  }
+
+  override def createAndRegisterDoubleHistogramWithNative[A](
       prefix: Option[Metric.Prefix],
       name: Histogram.Name,
       help: Metric.Help,
       commonLabels: Metric.CommonLabels,
       labelNames: IndexedSeq[Label.Name],
+      buckets: NonEmptySeq[Double],
       nativeHistogram: NativeHistogram
```
hgdyeo: **[Major]** Override params still named `nativeHistogram: NativeHistogram` — inconsistent with PR #129's SPI rename to `config: NativeHistogram` ([discussion](https://github.com/permutive-engineering/prometheus4cats/pull/129#discussion_r3275531265)).

Applies in two places:
- `:291` — `createAndRegisterDoubleHistogramWithNative` (dual-mode)
- `:354` — `createAndRegisterDoubleNativeHistogram` (native-only)

Plus the ~10 in-body references (`nativeHistogram.initialSchema`, `.maxNumberOfBuckets`, etc.) in each method.

Scala doesn't require override param names to match the trait, but reading the override against the trait is confusing when they differ. Mechanical rename — no behaviour change.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -233,29 +238,174 @@ class JavaMetricRegistry[F[_]: Async] private (
...
+      Histogram.make[F, Double, A](
+        Histogram.ExemplarState.fromRef(buckets, exemplarRef),
+        _observe = { (d: Double, labels: A, exemplar: Option[Exemplar.Labels]) =>
+          Utils.modifyMetric[F, Histogram.Name, io.prometheus.metrics.core.datapoints.DistributionDataPoint](
+            metricName = name,
+            allLabelNames = allLabelNames,
+            dynamicLabels = f(labels),
+            commonLabelValues = commonLabelValuesArray,
+            getDataPoint = (lbls: Array[String]) => histogram.labelValues(lbls: _*),
+            modify = (dp: io.prometheus.metrics.core.datapoints.DistributionDataPoint) =>
```
hgdyeo: **[Nice-to-have]** The three histogram methods (`createAndRegisterDoubleHistogram` at `:237`, `createAndRegisterDoubleHistogramWithNative` at `:285`, `createAndRegisterDoubleNativeHistogram` at `:348`) share ~25 lines of identical post-registration plumbing each — the `commonLabels` extraction prologue and the `Histogram.make[F, Double, A] { _observe = ... Utils.modifyMetric(...) }` wrapper. Only the `Histogram.Builder` chain and `ExemplarState` argument actually differ between modes.

Could extract a private helper:

```scala
private def histogramFromBuilder[A](
    prefix: Option[Metric.Prefix],
    name: Histogram.Name,
    commonLabels: Metric.CommonLabels,
    labelNames: IndexedSeq[Label.Name],
    metricType: MetricType,
    buildBuilder: () => PHistogram,
    exemplarState: Ref[F, Option[Exemplar.Data]] => Histogram.ExemplarState
)(f: A => IndexedSeq[String]): Resource[F, Histogram[F, Double, A]]
```

Trade-off: collapsing ~75 lines of plumbing into one helper + three callers (~10 lines each) means one level of indirection to trace each registration path. The current side-by-side form makes it easy to grep-diff the three modes — extracting hides that contrast.

No strong opinion. Mostly worth a deliberate decision either way. Push back if you'd rather keep the explicit three-method form.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -233,29 +238,174 @@ class JavaMetricRegistry[F[_]: Async] private (
...
+    val fullName               = NameUtils.makeName(prefix, name)
 
-  override def createAndRegisterDoubleNativeHistogram[A](
+    configureBuilderOrRetrieve[PHistogram](
+      register = () =>
+        PHistogram
+          .builder()
+          .name(fullName)
+          .help(help.value)
+          .labelNames(allLabelNames.map(_.value): _*)
```
hgdyeo: **[Nice-to-have]** `buckets.toSeq.toArray: _*` materialises a Scala `Seq` and then a primitive `double[]` per registration — two intermediate collections.

`buckets.toList: _*` would work identically (the Scala compiler inserts the conversion to `double[]` for the Java varargs call thanks to the `ClassTag[Double]` in scope). Single intermediate, cleaner read. The `.toSeq.toArray` form looks defensive against an imagined boxing risk that Scala's varargs ascription doesn't actually have for primitive-element collections.

Applies in two places:
- `:253` (classic) — `.classicUpperBounds(buckets.toSeq.toArray: _*)`
- `:301` (dual-mode) — same

Registration is a one-off, so any allocation pattern is fine at runtime. Suggestion is purely about readability — push back if you'd rather keep the explicit `.toArray` materialisation.

hgdyeo:
Going to fix all three as I think they are fairly uncontroversial

hgdyeo:
[b79d42d](https://github.com/permutive-engineering/prometheus4cats/pull/133/commits/b79d42d73bb81e289f24558b5bac48429c1c7cb6) meh, a bit marginal

PR #134, opened 2026-05-21
Implement Summary and Info in javaclient

hgdyeo:
Replaces the last two `notYetPorted` registration stubs in the `javaclient` backend (Summary and Info) with real implementations against `prometheus-metrics-core` 1.x.

After this PR: every non-callback metric-registration method on the new backend is implemented. The only remaining stubs are the five callback registrations + the `MetricCollection` callback, which land in PR 9.

## Summary impl

Builds `PSummary` via the upstream builder, applying:
- `.maxAgeSeconds(maxAge.toSeconds)` + `.numberOfAgeBuckets(ageBuckets.value)`
- Each declared `Summary.QuantileDefinition` mapped to `.quantile(q.value, q.error)` via a `quantiles.foreach` side-effect loop (the upstream builder mutates in-place and returns self for chaining; the loop is the idiomatic way to apply N quantile setters).
- Registration via the shared `configureBuilderOrRetrieve` helper for ref-counted registration semantics.

## Info impl

Uses upstream's `Info.builder().labelNames(...).register()` then `setLabelValues(values...)` at observation, matching PR #130's redesigned SPI (`Info[F, A]` with build-time labelNames and curried label-value projection).

Uses an **inline state machine** for registration rather than `configureBuilderOrRetrieve`, because Info extends `MetricWithFixedMetadata` directly (not `StatefulMetric`) and doesn't need exemplar tracking. The inline machine mirrors the same ref-counted registration + cross-axis collision detection semantics.

Subtle: upstream's `Info.builder()` strips the `_info` suffix internally and re-appends it during exposition. The user-supplied `Info.Name` is constrained to end in `_info` (regex on the refined Name type), so this adapter passes the raw name through and lets upstream handle the round-trip.

## Context

PR 8 of the v6 backend uplift split out of #124. Stacked on #133 (javaclient histograms).

Two source commits combined:
- `23c94adfb2c82e6f5eeef0503bf8789f2eb46954` — Summary impl. Conflict resolution carried over PR #133's helper-extraction refactor's Summary prologue while keeping the cherry-picked impl body, plus replaced the deferred v5-shape Info stub with the v6 `[A]`-shape from PR #132's SPI alignment.
- `javaclient/` hunks from `b14e0bbaa758d0888230370e227d540f4e502892` — Info impl + the new `+ Info` smoke test. The core SPI redesign hunks already landed in PR #130.

## Risk

Low — pure extension over PR 133's backend. Two new method bodies replacing throwing stubs. No SPI changes (committed in PR #129/#130/#131).

## Test plan

- [x] `sbt ci-test` green locally on Scala 2.13.18 + 3.3.7 (123 tests pass, +2 new Info smoke tests vs PR 133's 121).

## Reviewer focus

- **Quantile side-effect loop** (`quantiles.foreach(q => builder.quantile(q.value.value, q.error.value))`) — `PSummary.Builder.quantile(...)` mutates and returns self. The `.foreach` form is shorter than fold-style, intentional.
- **Info's inline state machine** vs the shared `configureBuilderOrRetrieve` — different because Info is `MetricWithFixedMetadata`, not `StatefulMetric`, and skips the exemplar Ref machinery. State-key/dedup/refcount semantics are identical to the shared helper.
- **The `_info` suffix round-trip** — `Info.Name`'s regex enforces the `_info` suffix; upstream strips and re-appends. Worth confirming the smoke test asserts on the exposition with the suffix.
- **`@SuppressWarnings(Array("scalafix:DisableSyntax.=="))` on the Info method** — same scalafix-rule trap as `configureBuilderOrRetrieve`. The `metricId == expected` check is a same-type tuple comparison; suppression is targeted to this one method.

### DIFF
--- a/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
+++ b/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
@@ -45,6 +45,7 @@ import io.prometheus.metrics.core.metrics.{Summary => PSummary}
 import io.prometheus.metrics.instrumentation.jvm.JvmMetrics
 import io.prometheus.metrics.model.registry.PrometheusRegistry
 import io.prometheus.metrics.model.registry.{Collector => PCollector}
+import io.prometheus.metrics.model.registry.{MultiCollector => PMultiCollector}
 import io.prometheus.metrics.model.snapshots.ClassicHistogramBuckets
 import io.prometheus.metrics.model.snapshots.CounterSnapshot
 import io.prometheus.metrics.model.snapshots.DataPointSnapshot
@@ -54,6 +55,7 @@ import io.prometheus.metrics.model.snapshots.HistogramSnapshot
 import io.prometheus.metrics.model.snapshots.Labels
 import io.prometheus.metrics.model.snapshots.MetricMetadata
 import io.prometheus.metrics.model.snapshots.MetricSnapshot
+import io.prometheus.metrics.model.snapshots.MetricSnapshots
 import io.prometheus.metrics.model.snapshots.Quantiles
 import io.prometheus.metrics.model.snapshots.SummarySnapshot
 import io.prometheus.metrics.model.snapshots.{Quantile => PQuantile}
@@ -83,6 +85,20 @@ class JavaMetricRegistry[F[_]: Async] private (
     private val dispatcher: Dispatcher[F],
     private val callbackTimeout: FiniteDuration,
     private val singleCallbackTimeout: FiniteDuration,
+    // Metric-collection callbacks: keyed by prefix (each prefix has its own merged common-labels map
+    // and a token-keyed map of user callbacks producing MetricCollection values). Constructed
+    // here, but the upstream MultiCollector that aggregates them is built in the class body via
+    // [[buildMetricCollectionMultiCollector]] so it can capture `this`.
+    private val metricCollectionRef: Ref[F, Map[
+      Option[Metric.Prefix],
+      (
+          Map[Label.Name, String],
+          Map[
+            Unique.Token,
+            F[MetricCollection]
+          ]
+      )
+    ]],
     private val logger: Throwable => String => F[Unit]
 ) extends DoubleMetricRegistry[F]
     with DoubleCallbackRegistry[F] {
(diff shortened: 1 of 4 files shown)


hgdyeo (commented):
Seven review comments — one Major (the Info NOTE contradicts upstream source), one Discussion (the `_info`/`_total` suffix-strip across Info and Counter — multiple options, no prescription), three Nice-to-have, one FYI-but-flagged on the catch-all release semantics, one FYI on the State type widening. Inline below.

modules/prometheus4cats-java/src/test/scala/prometheus4cats/javaclient/JavaMetricRegistrySuite.scala
```diff
@@ -189,6 +191,67 @@ class JavaMetricRegistrySuite extends CatsEffectSuite {
...
+            val dp       = snapshot.getDataPoints.asScala.head
+            val labelMap = dp.getLabels.asScala.map(l => l.getName -> l.getValue).toMap
+            assertEquals(labelMap.get("version"), Some("1.2.3"))
+            assertEquals(labelMap.get("commit"), Some("abc1234"))
+          }
+        }
+    }
+  }
+
+  // NOTE: An "Info with no labels declared" test was attempted and removed — upstream
```
hgdyeo, line 249: **[Major]** This NOTE contradicts upstream source. The claim is that "Info with empty labelNames produces no observable scrape entry" — but [`Info.java:collect()`](https://github.com/prometheus/client_java/blob/main/prometheus-metrics-core/src/main/java/io/prometheus/metrics/core/metrics/Info.java) does:

```java
if (labelNames.length == 0) {
  data.add(new InfoSnapshot.InfoDataPointSnapshot(constLabels));
}
```

— which **does** emit a data point when `labelNames` is empty. The two statements can't both be true.

Most plausible explanation: the removed test searched for the snapshot under the wrong name. Upstream's `Info.Builder.name(...)` calls `stripInfoSuffix(name)`, so an Info registered as `test_info` is stored under `test`. A test filtering on `getMetadata.getName === "test_info"` would find nothing → conclude "no scrape entry" → not the same as "Info emits nothing".

**Suggested:** restore the attempted no-label-Info test with corrected name matching (e.g., filter on `getName === "test"` and assert a single `InfoSnapshot.InfoDataPointSnapshot`), or delete the misleading NOTE entirely. The current state risks feeding wrong guidance into the v6 migration guide (PR 11).

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -385,16 +387,126 @@ class JavaMetricRegistry[F[_]: Async] private (
...
-    Resource.eval(ApplicativeThrow[F].raiseError(notYetPorted("createAndRegisterInfo")))
+  )(f: A => IndexedSeq[String]): Resource[F, Info[F, A]] = {
+    // Upstream client_java appends "_info" to the metric name automatically when the registered
+    // name doesn't end in "_info". The user-provided Info.Name is constrained to end in "_info"
+    // already (see Info.Name regex), so we strip the suffix before registering and let upstream
+    // re-add it.
+    val rawName  = name.value
+    val baseName = if (rawName.endsWith("_info")) rawName.dropRight("_info".length) else rawName
+
+    // Info uses MetricWithFixedMetadata (not StatefulMetric), so the registration state machinery
```
hgdyeo, line 491: **[Discussion]** Several intertwined points on the `_info` suffix-strip — worth a maintainer decision, no single prescription from me.

**(a) The scaladoc is misleading.** It says "we strip the suffix before registering and let upstream re-add it" — but the code passes `rawName` (with `_info`) to `.name()` at line 478. Upstream's `Info.Builder.name(...)` does the strip internally. The Scala-side strip is for our **local State-map key only**, not for the registration call.

Actual flow:
- `rawName` → upstream's `.name()` → upstream strips, stores under base name internally.
- `baseName` → our State-map key (matches upstream's internal stored name).
- `registry.unregister(collector)` uses the collector reference, not a name, so the State key is never cross-checked.

**(b) The `else rawName` branch is dead code.** `Info.Name`'s regex (`^[a-zA-Z_:][a-zA-Z0-9_:]*_info$`) guarantees the suffix; `rawName.endsWith("_info")` is always true. Either drop the conditional or document the defensive intent against future regex changes.

**(c) Functionally unnecessary.** Every `Info.Name` ends in `_info`, so the mapping `rawName → baseName` is deterministic 1:1. No two `Info.Name` values can map to the same baseName but differ in rawName — using `baseName` vs `rawName` as the State key is two equivalent encodings of the same identity.

**(d) Counter does the same thing, with two inconsistencies.** `counterName` at line 71 strips `_total` for the same purpose (legacy adapter does the same — `infoNameShow` Show instance and `counterName` helper). Two small mechanic differences:
- Counter uses `.replace("_total", "")` (strips ALL occurrences); Info uses `.dropRight("_info".length)` (strips trailing only). `Counter.Name("foo_total_bar_total")` → Counter strips both → `"foo_bar"`, possibly unintended. In practice degenerate names are unlikely, but the inconsistency exists.
- Counter has a helper at the class level; Info inlines the same logic. No structural reason for the difference.

**(e) Why is this here at all?** Three plausible reasons, in order of likelihood:
1. Inherited from the legacy `javasimpleclient` adapter (`counterName` + `infoNameShow` Show instance, line 91 + 380 of legacy) without re-evaluation. Likely predates the regex-enforced `_total$` / `_info$` requirement.
2. Aesthetic alignment with upstream's internal storage (`stripTotalSuffix`, `stripInfoSuffix` in 1.x). Symmetric but never cross-referenced.
3. Defensive against future regex relaxation.

## Options

**A) Keep the strip; fix only the surface issues.** Rewrite the misleading Info scaladoc, drop the dead `else` branch, optionally fix Counter's `.replace` → `.dropRight` / `.stripSuffix` for the degenerate-name correctness. Cheapest, preserves established pattern.

**B) Simplify Info to use `rawName` throughout; leave Counter alone.** Drops the strip + dual name variables + dead branch in Info. Improves error messages (consumer sees `'build_info'` matching what they typed). Creates a small Counter/Info asymmetry.

**C) Simplify both to use raw name throughout.** Consistent. Removes the helper + inline strip + both dead branches. Largest blast radius (touches Counter).

**D) Keep the code as-is, but document accurately in both places.** Replace the misleading Info scaladoc + add an equivalent comment to `counterName` saying "this aligns our State key with upstream's internal stored name; the regex makes it functionally unnecessary, kept for symmetry with the legacy adapter pattern."

Factual claims above (regex enforcement, control flow, `.replace` vs `.dropRight` mechanic) are all verifiable. Whichever way the team wants to go.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -385,16 +387,126 @@ class JavaMetricRegistry[F[_]: Async] private (
...
+        ref.get.flatMap { metrics =>
+          metrics.get(fullName) match {
+            case Some((`metricId`, (_, _, 1))) =>
+              ref.set(metrics - fullName) >>
+                Sync[F].delay(registry.unregister(collector)).handleErrorWith { e =>
+                  logger(e)(s"Failed to unregister Info collector: '$collector'")
+                }
+            case Some((`metricId`, (collector, exemplarRef, references))) =>
+              ref.set(metrics.updated(fullName, (metricId, (collector, exemplarRef, references - 1))))
+            case _ => Applicative[F].unit
```
hgdyeo: **[FYI but flagged]** Catch-all `case _ => Applicative[F].unit` here (and at line 132 in `configureBuilderOrRetrieve`'s release) silently swallows unexpected state.

The catch-all matches two scenarios:
- **`None`** — the State entry was already removed. Possible if release runs twice; Cats Effect `Resource.use` guarantees release-once-per-acquire, so shouldn't happen normally.
- **`Some((differentMetricId, ...))`** — same State key but different metricId. Could only happen if the State entry was replaced between our acquire and release; the semaphore-guarded acquire/release loop *should* make this impossible.

Both cases require a bug elsewhere in the state machinery. In practice they shouldn't fire.

**Implication if either case DOES fire:** the collector handed to release is **not unregistered** from upstream's `PrometheusRegistry`. Our local State stays consistent — but upstream's registry holds an orphaned collector that nothing tracks. The orphan continues to appear in scrape output indefinitely (until JVM dies) and may collide with a future registration of the same name, raising a confusing error far from the original bug. Slow leak that's invisible at the API boundary.

## Options

1. **Status quo** — accept the catch-all as defensive against impossible states. Silent upstream-registry leak if a bug ever surfaces.
2. **Log on the catch-all** — `logger(...)` with a clear message. Cheap; surfaces the leak as a log line. Doesn't fix the orphan but makes it observable.
3. **Try to unregister anyway** — even on the catch-all, call `registry.unregister(collector)`. Prevents the orphan but risks unregistering something that's actually still in use elsewhere.

The same pattern lives in `configureBuilderOrRetrieve` (line 132) — inherited from the legacy adapter. Whichever option the team picks, consistency suggests applying it to both places.

Bugs leading to this are unlikely in practice; flagging for completeness so the latent risk is a deliberate decision rather than an unexamined inheritance.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -385,16 +387,126 @@ class JavaMetricRegistry[F[_]: Async] private (
...
+          }
+        }
```
hgdyeo, line 552: **[Nice-to-have]** The Info impl's inline state machine (lines 449-498) is structurally identical to `configureBuilderOrRetrieve` (lines 84-137):
- Same `sem.permit.surround(ref.get.flatMap { metrics => metrics.get(fullName) match { ... } })` shape.
- Same dedup via `metricId == expected` (with the same targeted `@SuppressWarnings(Array("scalafix:DisableSyntax.=="))`).
- Same `Resource.make`/release with refcount decrement + unregister.

~50 lines of similar logic. The reason given (correctly) is that `configureBuilderOrRetrieve[M <: StatefulMetric[_, _]]` can't accept `PInfo` (which extends `MetricWithFixedMetadata` directly, not `StatefulMetric`), so the helper had to be either widened or duplicated for Info.

## Options

**A)** Widen `configureBuilderOrRetrieve` to `M <: MetricWithFixedMetadata`. The State already widened to `MetricWithFixedMetadata` in `package.scala`, so this is mostly a constraint-relaxation. The exemplar-Ref machinery becomes "always create one even if Info ignores it" — minor wasted Ref per Info registration. ~30 lines saved.

**B)** Extract the inner state-machine block into a private `def withRegistration[M <: MetricWithFixedMetadata](...)` and have both `configureBuilderOrRetrieve` and the Info impl call into it. More architecturally honest about Info-vs-the-others.

**C)** Accept the duplication. Both impls are clean on their own; the duplication isn't actively confusing.

No strong opinion. Removing it would be a one-shot 30-50 line win; keeping it costs nothing but lines.

modules/prometheus4cats-java/src/test/scala/prometheus4cats/javaclient/JavaMetricRegistrySuite.scala
```diff
@@ -189,6 +191,67 @@ class JavaMetricRegistrySuite extends CatsEffectSuite {
...
   }
 
+  test("summary — quantiles propagate and observations are aggregated") {
+    freshRegistry.use { case (promRegistry, factory) =>
+      factory
+        .summary("test_summary")
+        .ofDouble
+        .help("test summary")
+        .quantile(Summary.Quantile.from(0.5).toOption.get, Summary.AllowedError.from(0.05).toOption.get)
+        .quantile(Summary.Quantile.from(0.99).toOption.get, Summary.AllowedError.from(0.01).toOption.get)
```
hgdyeo, line 202: **[Nice-to-have]** `.toOption.get` × 4 here for refined-type construction of constant, known-valid values (`0.5`, `0.99`, `0.05`, `0.01`).

The `.toOption.get` pattern is a smell — explicit `.get` on `Option`. Cleaner alternatives:
- `Summary.Quantile.unsafeFromDouble(0.5)` if such an unsafe constructor exists.
- `Summary.Quantile.from(0.5).getOrElse(throw new IllegalArgumentException("compile-time constant"))` — more explicit failure mode.
- Or extract small `q(...)` / `err(...)` helpers at the top of the test file to centralise the smell.

Low priority — test code, known-valid values. Flagging because the project style avoids `.get` on `Option`.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -385,16 +387,126 @@ class JavaMetricRegistry[F[_]: Async] private (
...
+      labels = allLabelNames
+    ).map { case (summary, _) =>
+      Summary.make[F, Double, A] { case (d, labels) =>
+        Utils.modifyMetric[F, Summary.Name, io.prometheus.metrics.core.datapoints.DistributionDataPoint](
+          metricName = name,
+          allLabelNames = allLabelNames,
+          dynamicLabels = f(labels),
+          commonLabelValues = commonLabelValuesArray,
+          getDataPoint = (lbls: Array[String]) => summary.labelValues(lbls: _*),
+          modify = (dp: io.prometheus.metrics.core.datapoints.DistributionDataPoint) => dp.observe(d),
```
hgdyeo: **[Nit]** Long fully-qualified type names like `io.prometheus.metrics.core.datapoints.DistributionDataPoint` are inlined throughout the file instead of imported. Affects multiple sites — Counter/Gauge/Histogram/Summary/Info all reference `CounterDataPoint`, `DistributionDataPoint`, `GaugeDataPoint` inline.

**Suggested:** add a single import at the top of the file:

```scala
import io.prometheus.metrics.core.datapoints.{CounterDataPoint, DistributionDataPoint, GaugeDataPoint}
```

(or `import io.prometheus.metrics.core.datapoints._` for the wildcard form). Then bare type names throughout. Mechanical change; an IntelliJ "Optimise Imports" or single global find-replace will handle it.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/package.scala
```diff
@@ -29,8 +29,12 @@ package object javaclient {
 
   private[javaclient] type MetricID = (IndexedSeq[Label.Name], MetricType)
 
+  /** State entry bound by the most specific common parent of every upstream metric type we register —
+    * `MetricWithFixedMetadata`. Counter, Gauge, Histogram, Summary all extend `StatefulMetric` which extends this; Info
+    * extends this directly without going through `StatefulMetric`.
+    */
   private[javaclient] type StateValue[F[_]] =
-    (MetricID, (StatefulMetric[_, _], Ref[F, Option[Exemplar.Data]], Int))
+    (MetricID, (MetricWithFixedMetadata, Ref[F, Option[Exemplar.Data]], Int))
```
hgdyeo, line 41: **[FYI]** `StateValue` widened from `StatefulMetric[_, _]` → `MetricWithFixedMetadata`. Notable cross-cutting structural change — the registry's internal map now stores collectors at the lowest common ancestor of all five metric types.

Consequences:
- `configureBuilderOrRetrieve`'s `collector.asInstanceOf[M]` cast still works (M is bound by `StatefulMetric[_, _]` which extends `MetricWithFixedMetadata`).
- The Info impl can store `PInfo` (which extends `MetricWithFixedMetadata` directly) in the same map.
- Cross-type dedup via `MetricType.Info` still distinguishes Info from the four StatefulMetric subtypes, so the "same name as Counter and Info" case correctly raises.

The in-file scaladoc already covers this. Flagging in the review for visibility in the PR description — worth surfacing as a deliberate cross-cutting change so it's not buried.

modules/prometheus4cats-java/src/test/scala/prometheus4cats/javaclient/JavaMetricRegistrySuite.scala
hgdyeo: I vote "will not fix" on this one

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
hgdyeo: will fix

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
hgdyeo: I vote "will not fix"

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
hgdyeo: 2. seems reasonable to me

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
hgdyeo: Addressed in 1a5aa07 — added `import io.prometheus.metrics.core.datapoints.{CounterDataPoint, DistributionDataPoint, GaugeDataPoint}` (split into 3 single-import lines by scalafmt) at the top, replaced all 6 inline FQ references with bare names.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
hgdyeo: Addressed in 053ce69 (option 2). Both catch-all branches (`configureBuilderOrRetrieve` release at the helper level, and the inline Info release) now log via `logger(new IllegalStateException(...))("...")` with a 'please report' message naming the metric. Behaviour unchanged for the happy path; the latent orphan-leak is now observable in logs if it ever fires.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
hgdyeo: Acked — won't fix. Accepting the duplication; both impls are clean on their own and the duplication isn't actively confusing. Leaving option A/B as a potential follow-up if a future maintainer feels strongly.

modules/prometheus4cats-java/src/test/scala/prometheus4cats/javaclient/JavaMetricRegistrySuite.scala
hgdyeo: Acked — won't fix. Test code with known-valid constants; the smell is acceptable here.

hgdyeo: I vote "Will not fix"

PR #135, opened 2026-05-22
Implement callback machinery in javaclient (Counter, Gauge, Histogram, Summary, MetricCollection)

hgdyeo:
Final piece of the javaclient backend: implements all five callback registration methods. After this PR every `MetricRegistry` + `CallbackRegistry` abstract method on the new backend has a real implementation. Ready for the testkit conformance wire-up in PR 10.

## Summary

Four cherry-picked commits from #124 + one local lint fix:

- **`fb04823` Add callback machinery to javaclient backend; implement Counter callback** — introduces the shared `CallbackPayload` / `CallbackState` types, the `registerCallback` private helper, callback-timeout state Refs, the Dispatcher wiring, and the first concrete callback (`registerDoubleCounterCallback`). The bulk of the structural work.
- **`ac6e256` Implement Gauge callback in javaclient backend** — second callback, mostly mechanical.
- **`46059df` Implement Histogram and Summary callbacks in javaclient backend** — refactors `CallbackPayload` to `Map[Token, F[NEL[DataPointSnapshot]]]` (the upstream parent type) so all four kinds share one storage; each callback's snapshot construction happens inside the user effect; the kind-specific collector casts back at scrape time. Three smoke tests added: Histogram callback exposes `Histogram.Value` shape, Summary callback exposes `Summary.Value` shape, plus the existing Counter + Gauge tests.
- **`31a8925` Implement registerMetricCollectionCallback in javaclient backend** — the heaviest of the five. Builds a single `PCollector` that aggregates ALL declared callbacks under the same prefix + commonLabels, with combined-timeout + per-callback-timeout error reporting via `callbackErrorState` and `singleCallbackErrorState` Refs.
- **`01089e5` Suppress DisableSyntax.null on callback methods + drop unused s-interpolator** — local lint fix. Upstream's Java constructors take `@Nullable Exemplar` and we pass `null` as the sentinel for "no exemplar"; scalafix's `DisableSyntax.null` rule flagged each site. Targeted suppressions on the four affected methods (consistent with PR #132's `DisableSyntax.==` pattern).

Diff: 3 files, ~940 lines added (including 5 new smoke tests). Tests: 128 passing × Scala 2.13.18 + 3.3.7 (was 123 — five new callback tests).

## Carve note

Commit 12's `MetricFactory.scala` hunks (the `HistogramWithCallbacksAndNativeDsl` plumbing fix) already landed in PR #131 (back-ported there to resolve mdoc CI failure). The conflict was resolved by keeping the version from `series/6.x` (which uses the fully-qualified `internal.HistogramMetricDsl.WithCallbacksImpl` scaladoc link from #132's follow-up fix).

## Context

PR 9 of the v6 backend uplift split out of #124. Stacked on #134 (javaclient Summary + Info).

After PR 9 + PR 10 (testkit conformance + delete legacy `javasimpleclient`) + PR 11 (docs), the v6 series should be tag-ready.

## Risk

Medium-to-high (largest single PR by code volume). The callback machinery is genuinely subtle:

- Five concurrent registration paths sharing one `CallbackState` Ref keyed by `Token`.
- Two timeout strategies in play: per-callback (`singleCallbackTimeout`) and aggregate (`callbackTimeout`).
- Error state is partitioned into "timed out" vs "failed" Refs to avoid log spam on repeat scrapes.
- Snapshot construction happens **inside the user's `F`** during the upstream registry's scrape thread (via Dispatcher.unsafeRunSync), so cancellation + thread-safety are important.

The smoke tests cover the happy path for each callback kind end-to-end through a real `PrometheusRegistry.scrape()`.

## Test plan

- [x] `sbt ci-test` green locally on Scala 2.13.18 + 3.3.7 (128 tests).
- [ ] CI green on JDK 11 / 17 / 21 / 25.
- [ ] Testkit conformance suite (lands in PR 10) will exercise these paths comprehensively.

## Reviewer focus

- **`registerCallback` helper** in the new commit `fb04823` — the shared registration shape for all five callbacks. Three Refs interact: `callbackState` (the actual payload map), `callbackErrorState`/`callbackTimeoutState` for aggregate-level error tracking, and `singleCallbackErrorState` for per-callback dedup.
- **`runCallbackWithBounds` / `timeoutEachCallback`** — the two timeout strategies. The aggregate timeout wraps the whole batch; the per-callback timeout wraps each user callback individually so one slow callback doesn't poison the batch.
- **Dispatcher usage** — required because upstream's `Collector.collect()` is a synchronous Java method but we need to call `F[NEL[DataPointSnapshot]]` from it. `dispatcher.unsafeRunTimed(...)` bridges the two worlds.
- **`metricCollectionToSnapshots`** in commit `31a8925` — the heaviest method in the file (~150 lines). Maps prometheus4cats `MetricCollection` → upstream `MetricSnapshots`. Each metric kind has its own per-value mapping.
- **`asInstanceOf[*DataPointSnapshot]` casts** — safe because the public `register*Callback` signatures constrain the input type at the API boundary; internal storage uses the parent type so all four kinds share one `CallbackState`.
- **The four `null: Exemplar` sites** — passing null to upstream's `@Nullable Exemplar` constructor parameter. Standard Java interop; suppressed via `@SuppressWarnings(Array("scalafix:DisableSyntax.null"))` on the enclosing methods.

### DIFF
--- a/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
+++ b/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
@@ -45,6 +45,7 @@ import io.prometheus.metrics.core.metrics.{Summary => PSummary}
 import io.prometheus.metrics.instrumentation.jvm.JvmMetrics
 import io.prometheus.metrics.model.registry.PrometheusRegistry
 import io.prometheus.metrics.model.registry.{Collector => PCollector}
+import io.prometheus.metrics.model.registry.{MultiCollector => PMultiCollector}
 import io.prometheus.metrics.model.snapshots.ClassicHistogramBuckets
 import io.prometheus.metrics.model.snapshots.CounterSnapshot
 import io.prometheus.metrics.model.snapshots.DataPointSnapshot
@@ -54,6 +55,7 @@ import io.prometheus.metrics.model.snapshots.HistogramSnapshot
 import io.prometheus.metrics.model.snapshots.Labels
 import io.prometheus.metrics.model.snapshots.MetricMetadata
 import io.prometheus.metrics.model.snapshots.MetricSnapshot
+import io.prometheus.metrics.model.snapshots.MetricSnapshots
 import io.prometheus.metrics.model.snapshots.Quantiles
 import io.prometheus.metrics.model.snapshots.SummarySnapshot
 import io.prometheus.metrics.model.snapshots.{Quantile => PQuantile}
@@ -83,6 +85,20 @@ class JavaMetricRegistry[F[_]: Async] private (
     private val dispatcher: Dispatcher[F],
     private val callbackTimeout: FiniteDuration,
     private val singleCallbackTimeout: FiniteDuration,
+    // Metric-collection callbacks: keyed by prefix (each prefix has its own merged common-labels map
+    // and a token-keyed map of user callbacks producing MetricCollection values). Constructed
+    // here, but the upstream MultiCollector that aggregates them is built in the class body via
+    // [[buildMetricCollectionMultiCollector]] so it can capture `this`.
+    private val metricCollectionRef: Ref[F, Map[
+      Option[Metric.Prefix],
+      (
+          Map[Label.Name, String],
+          Map[
+            Unique.Token,
+            F[MetricCollection]
+          ]
+      )
+    ]],
     private val logger: Throwable => String => F[Unit]
 ) extends DoubleMetricRegistry[F]
     with DoubleCallbackRegistry[F] {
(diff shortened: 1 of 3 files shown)


modules/prometheus4cats-java/src/test/scala/prometheus4cats/javaclient/JavaMetricRegistrySuite.scala
```diff
@@ -252,6 +253,204 @@ class JavaMetricRegistrySuite extends CatsEffectSuite {
   // Info almost always carries identity labels like version/commit/instance), but should be
   // surfaced in the v6 migration guide for any consumer that relied on a no-label `Info[F, Unit]`.
 
+  test("counter callback — scrape invokes the user callback and propagates the value to the snapshot") {
```
hgdyeo, line 256: **[Major] Test gap — timeout / error / dedup machinery is untested.**

The 5 new smoke tests (this one + the next four) cover the happy path only. The substantial machinery added in `JavaMetricRegistry.scala:580-777` ships untested:

- per-callback timeout (`singleCallbackTimeout`)
- aggregate timeout (`callbackTimeout`)
- callback `F.raiseError`
- the three error-state Refs (`callbackTimeoutState`, `callbackErrorState`, `singleCallbackErrorState`) deduping correctly

Deferring to a follow-up PR is fine — flagging so it doesn't get lost.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -557,21 +943,244 @@ class JavaMetricRegistry[F[_]: Async] private (
...
+      val result: F[MetricSnapshots] = metricCollectionRef.get.flatMap { perPrefix =>
+        // For each prefix, run all registered callbacks under the per-callback timeout, merge their
+        // MetricCollection values, then convert to MetricSnapshots. We use a fresh
+        // `singleCallbackErrorState` reading on each scrape so log-once gating still applies.
+        perPrefix.toList.flatTraverse { case (prefix, (commonLabels, callbacks)) =>
+          val mergedCol: F[MetricCollection] = callbacks.values.toList.foldM(MetricCollection.empty) { (acc, cbF) =>
+            cbF
+              .timeout(singleCallbackTimeout)
+              .handleErrorWith { th =>
+                logger(th)(
```
hgdyeo, line 1018: **[Major] `metricCollectionCollector` doesn't dedupe failure logs.**

Three log-emitting branches here (this line + 1025 + 1027) — none wrapped in `trackErrors`. Compare with `runAggregateCollect` (line 750) which dedupes via `callbackTimeoutState` / `callbackErrorState`.

Consequence: a persistently-failing `MetricCollection` callback logs **once per scrape** (~5760/day at 15s scrape interval). A failing Counter/Gauge/Histogram/Summary callback only logs once per process.

Fix is ~15 lines: add `metricCollectionErrorState: Ref[F, Set[String]]` and wrap the three log sites in `trackErrors`. A single sentinel key (e.g. `"<metric-collection-callback>"`) works since there's no per-metric name available.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -519,25 +560,324 @@ class JavaMetricRegistry[F[_]: Async] private (
...
+
+    // Build CounterDataPointSnapshot instances inside the user callback's F so the generic
+    // CallbackPayload type can hold them as plain DataPointSnapshots. The Collector at scrape time
+    // casts back to CounterDataPointSnapshot.
+    val projected: F[NonEmptyList[DataPointSnapshot]] = callback.map(
+      _.map { case (v, a) =>
+        new CounterSnapshot.CounterDataPointSnapshot(
+          if (v < 0) 0.0 else v,
+          Labels.of(allLabelNamesStr, (f(a) ++ commonLabelValuesArr).toArray),
+          null: io.prometheus.metrics.model.snapshots.Exemplar,
```
hgdyeo: **[NIT] Same `Exemplar` FQ-name nit as PR #134 (fixed in b79f6b0).**

`Exemplar` is shadowed by `import prometheus4cats._`, so the FQ name is needed for the type ascription on `null`. The alias pattern already used elsewhere in this file (`PCounter`, `PGauge`, `PHistogram`, `PSummary`, `PInfo`, `PCollector`, `PMultiCollector`, `PQuantile`) is the standard fix:

```scala
import io.prometheus.metrics.model.snapshots.{Exemplar => PExemplar}
```

Then `null: PExemplar` at the four sites: this one (804), 853, 1096, 1115. The `@SuppressWarnings` annotations stay (the scalafix rule fires on `null`, not the type).

hgdyeo (commented):
Critical review — 7 comments. #1, #2 are major (test gap, log spam). #3 is a cheap consistency fix matching PR #134 #5. #4 is a should-fix. #5-#7 are smaller / FYI.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -57,7 +77,28 @@ import prometheus4cats.util.NameUtils
 class JavaMetricRegistry[F[_]: Async] private (
     private val registry: PrometheusRegistry,
     private val ref: Ref[F, State[F]],
+    private val callbackState: Ref[F, CallbackState[F]],
+    private val callbackTimeoutState: Ref[F, Set[String]],
+    private val callbackErrorState: Ref[F, Set[String]],
+    private val singleCallbackErrorState: Ref[F, (Set[String], Set[String])],
```
hgdyeo: **[Should-fix] Positional tuple — `_1` is timeout-logged names, `_2` is error-logged names.**

`Ref[F, (Set[String], Set[String])]` is positional; the compiler can't enforce which slot is which. `collectAllPayload` (lines 721-744) destructures by position via `case (loggedTimeout, loggedError) => ...` — a future refactor that swaps the assignments would silently invert dedup behaviour, and the smoke tests would never notice.

Compare with `callbackTimeoutState` and `callbackErrorState` on lines 81-82 — two separate `Ref[F, Set[String]]`s, role enforced by variable name.

Suggested: split this into `singleCallbackTimeoutState` + `singleCallbackErrorLogState` for consistency with the aggregate-level pattern. ~10 lines touched.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -519,25 +560,324 @@ class JavaMetricRegistry[F[_]: Async] private (
...
+    // exposition writer adds "_total" back to the wire format.
+    val baseName = if (rawName.endsWith("_total")) rawName.dropRight("_total".length) else rawName
+
+    // Build CounterDataPointSnapshot instances inside the user callback's F so the generic
+    // CallbackPayload type can hold them as plain DataPointSnapshots. The Collector at scrape time
+    // casts back to CounterDataPointSnapshot.
+    val projected: F[NonEmptyList[DataPointSnapshot]] = callback.map(
+      _.map { case (v, a) =>
+        new CounterSnapshot.CounterDataPointSnapshot(
+          if (v < 0) 0.0 else v,
```
hgdyeo, line 809: **[FYI] Silent clamp of negative Counter values to 0.0.**

This matches v5 `javasimpleclient` behaviour. Upstream v6's `CounterDataPointSnapshot.validate()` throws `IllegalArgumentException` on negative values, so the clamp prevents a scrape-time crash on misuse.

Leaning toward keeping it (v5→v6 migration parity is worth more than strictness here), but currently reads as "weird code". Suggested: a one-line comment at the clamp explaining the deliberate v5-parity intent. Same site at line 1094 in `metricCollectionToSnapshots`.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -557,21 +943,244 @@ class JavaMetricRegistry[F[_]: Async] private (
...
+  /** Convert a `MetricCollection` (the prometheus4cats per-prefix bag of typed metric values) into a flat list of
+    * upstream `MetricSnapshot`s. Each (name, labelNames) entry becomes one snapshot with one or more data points (one
+    * per List entry in the user's MetricCollection).
+    */
+  @SuppressWarnings(Array("scalafix:DisableSyntax.null"))
+  private def metricCollectionToSnapshots(
+      prefix: Option[Metric.Prefix],
+      commonLabels: Map[Label.Name, String],
+      mc: MetricCollection
+  ): List[MetricSnapshot] = {
```
hgdyeo, line 1081: **[Note] Per-kind blocks (counter/gauge/histogram/summary) here through line 1183 share the same outer shape** (per-name grouping → `headOption` → wrap in `{Kind}Snapshot`) but differ in per-kind value extraction.

Extractable to a generic helper, but each kind's specifics (Counter's clamp, Histogram's bucket handling, Summary's `Quantiles.EMPTY` short-circuit) read more cleanly inline than they would behind a 7-arg helper. Flagging as a deliberate-choice note — YAGNI applies, but worth a maintainer eyeball before accepting the duplication.

modules/prometheus4cats-java/src/test/scala/prometheus4cats/javaclient/JavaMetricRegistrySuite.scala
```diff
@@ -252,6 +253,204 @@ class JavaMetricRegistrySuite extends CatsEffectSuite {
...
+                names.contains("CounterSnapshot" -> "collection_counter"),
+                s"expected CounterSnapshot 'collection_counter'; saw $names"
+              )
+              assert(
+                names.contains("GaugeSnapshot" -> "collection_gauge"),
+                s"expected GaugeSnapshot 'collection_gauge'; saw $names"
+              )
+              val counter = snapshots.collectFirst {
+                case s: CounterSnapshot if s.getMetadata.getName === "collection_counter" => s
+              }.get
```
hgdyeo, line 447: **[Nit] `.get` on Option + `.head` on a possibly-empty Java list** (this line + 448).

Other smoke tests in this file use `.getOrElse(fail(...))` for the same lookup shape (e.g. lines 67-68, 87-88). Same nit as PR #134 #4 — likely "will not fix" for consistency with the previous decision, flagging for the record.

modules/prometheus4cats-java/src/test/scala/prometheus4cats/javaclient/JavaMetricRegistrySuite.scala
hgdyeo: I vote defer.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
hgdyeo: Correction on the dedup-key suggestion above — the single-sentinel idea was too coarse. There are actually three log sites at **two** different granularities:

- **Line 1011 (per-individual-callback)**: runs once per registered callback per scrape inside the `foldM` over `callbacks.values`. If 3 callbacks are registered and all fail differently on scrape 1, a single sentinel would silence 2 of them — information loss.
- **Lines 1025, 1027 (aggregate)**: the whole batch timed out / threw. Single entity.

The kind-specific path uses the **metric name** as discriminator (`singleCallbackErrorState` keyed by metric name; `callbackTimeoutState` / `callbackErrorState` same). MetricCollection has no metric name, so the natural discriminators are:

- Inner per-callback → `Unique.Token` (each registered callback has one; see `metricCollectionRef` shape on lines 92-101).
- Outer aggregate → just ONE entity. A `Boolean` flag is enough, or a singleton sentinel `Set`.

Right shape is closer to:

```scala
private val metricCollectionPerCallbackState: Ref[F, (Set[Unique.Token], Set[Unique.Token])],  // (timed-out, errored)
private val metricCollectionAggregateState:   Ref[F, (Boolean, Boolean)],                       // (timed-out, errored)
```

(With the `(Set, Set)` tuple shape inheriting the same positional-encoding smell flagged in [comment](https://github.com/permutive-engineering/prometheus4cats/pull/135#discussion_r3289281995) — if this is fixed, this should follow the same split-into-two-Refs treatment.)

Two options for actioning:

**(a) Inner per-token dedup + outer per-entity dedup.** Two extra Refs (or four if split per [here](https://github.com/permutive-engineering/prometheus4cats/pull/135#discussion_r3289281995)). Behaviour mirrors the kind-specific path exactly — distinct failing callbacks each log once per process, persistent failures stay silent. Fix size ~25-30 lines.

**(b) Coarser sentinel-style dedup.** A single `Ref[F, Set[String]]` with one sentinel covering all three branches. Cheaper (~15 lines) but loses ability to surface distinct-callback failures — 3 callbacks failing differently on the same scrape would log 1 line, not 3.

hgdyeo: Given the slightly bigger fix, I'm leaning less toward fixing this, but I don't know how important the deduping is in practise.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
hgdyeo: Should fix

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
hgdyeo: Should fix

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
hgdyeo: a clear comment is my preference, should not fix.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
hgdyeo: will not fix

modules/prometheus4cats-java/src/test/scala/prometheus4cats/javaclient/JavaMetricRegistrySuite.scala
hgdyeo: will not fix

kidonm (commented):
Overall LGTM ≈ 85%. The implementation itself reads well — `registerCallback` is the obvious shared shape, the dual-timeout split is sensible, and the `asInstanceOf` storage trick falls out of the public API constraints rather than papering over them. After walking the full stack (#129 → #135) I have a few higher-level concerns that I think are best aired together rather than scattered across inline comments. None blocking on their own, but together they suggest the migration story is not quite finished — would love to grab 30 minutes to align on concrete next steps before this lands.

**1. Higher-level / integration test coverage.** The new backend has solid smoke tests (13 in `JavaMetricRegistrySuite`) and the testkit conformance suites are queued for PR 10, so per-method coverage will land. What is missing is an integration suite that drives the backend against a *real* Prometheus instance — HTTP scrape endpoint, protobuf scrape negotiation, ingest verification. Especially relevant for native histograms, where the protobuf path is what unlocks the whole feature, and currently no test exercises it end-to-end. Not blocking for this PR, but worth putting on the roadmap — it would serve as the baseline when we make deeper cuts like these in future.

**2. Native + classic histogram with same name — should be covered by a test.** The new `MetricType` ADT distinguishes `Histogram` / `NativeHistogram` / `HistogramWithNative` at our dedup layer, but upstream'\''s `PrometheusRegistry.register()` rejects same-name registrations regardless of kind. Net effect should be "fails loudly at Resource acquisition," which is reasonable behaviour — but we do not have a test pinning it. A one-liner here would close the question and document the contract.

**3. Observability regression on the library'\''s own metrics — needs a release + migration guide and ideally a scalafix.** The legacy `javasimpleclient` backend emits ten internal `prometheus4cats_*` metrics for self-monitoring (`registered_metrics`, `registered_metric_claims`, `callback_total`, `collection_callback_duration`, etc.). The new javaclient backend emits zero. There is a code comment acknowledging two of the ten ("may add in a follow-up commit"); the other eight are not mentioned anywhere. Anyone running dashboards or alerts off these names will silently lose visibility on migration. This deserves:

   - A migration guide enumerating *every* removed/renamed/changed metric, label, and API surface across the v5→v6 break (not just the ten internal metrics — also the Info redesign, the histogram DSL changes, the `.withNative` callback drop we discussed on #131).
   - A scalafix rule for the mechanical bits (Info shape, `nativeHistogram` parameter renames, anything else with a clean rewrite).
   - An explicit decision on whether the ten self-monitoring metrics come back or get formally dropped — either is fine, but the silent disappearance is not.

Happy to chat through these — feels like 30 minutes would let us agree on what is in-scope for this stack vs. what slips to follow-ups. Either way, the engineering here is good; just want to make sure the migration story is as solid as the implementation.

PR #145, opened 2026-06-03
series/6.x - deprecate .ofLong and .ofDouble; remove callbacks

kidonm:
- **Deprecate .ofLong / .ofDouble — default to Double**
- **Drop callback support entirely**


kidonm:
Splitting into a stack of two smaller PRs for easier review.

PR #150, opened 2026-07-06
Memoize data-point resolution per label value

Oliver-Taylor:
## Problem

Every update of a labelled metric re-resolves the upstream `DataPoint`: the typed label value `A` is rendered to strings (`f(labels)` — e.g. `UUID.toString` per label), copied into an array, and `prometheus-metrics-core`'s `labelValues(...)` then allocates a sorted `Labels` object and hashes it to find the data point — all before the actual `inc`/`observe`.

Profiling a metric-heavy production service (audience-matching-internal, ~560 req/s with per-request counter updates) attributed **~15% of CPU samples** to this plumbing: `ConcurrentHashMap` lookups, `String.equals`/`hashCode`, and label rendering, dwarfing the `DoubleAdder.add` doing the real work.

## Change

`DataPointResolver` (new, `javaclient.internal`) memoizes `A → DataPoint` per registered metric:

- **Correctness**: data points are never removed while a collector is registered (the library exposes no `remove`/`reset`), so cached references cannot go stale. Failed resolutions are not cached (`computeIfAbsent` stores nothing when the mapping function throws).
- **Hot path**: one lock-free `computeIfAbsent` hit, then the underlying adder/observation. Label rendering, array building, and the upstream `Labels` allocation/sort happen once per distinct label value.
- **Bounded**: beyond 65,536 distinct label values the cache stops accepting entries and resolves directly — label cardinality that high is already pathological for Prometheus, and the cap bounds memory even if an `A` without value-based equality is used.
- **Error messages**: rendered label values for `UnhandledPrometheusException` are computed lazily, only on the error path.

All four labelled metric types (counter, gauge, histogram, summary) go through the same seam, so each gets the cache from one change in `Utils.modifyMetric` plus its call sites.

Covered by a new `DataPointResolverSuite` (single resolution per label value, distinct values, failure non-caching, cap behaviour) plus the existing cross-backend `DslSuite`. Cross-compiles on 2.13.18 and 3.3.7.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
+++ b/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
@@ -38,6 +38,7 @@ import io.prometheus.metrics.instrumentation.jvm.JvmMetrics
 import io.prometheus.metrics.model.registry.PrometheusRegistry
 import io.prometheus.metrics.model.snapshots.Labels
 import prometheus4cats._
+import prometheus4cats.javaclient.internal.DataPointResolver
 import prometheus4cats.javaclient.internal.Utils
 import prometheus4cats.javaclient.models.MetricType
 import prometheus4cats.util.DoubleMetricRegistry
@@ -172,6 +173,11 @@ class JavaMetricRegistry[F[_]: Async] private (
       renderedName = fullName,
       labels = allLabelNames
     ).map { case (counter, exemplarRef) =>
+      val resolver = new DataPointResolver[A, CounterDataPoint](
+        f,
+        commonLabelValuesArray,
+        (lbls: Array[String]) => counter.labelValues(lbls: _*)
+      )
       Counter.make(
         Counter.ExemplarState.fromRef(exemplarRef),
         1.0,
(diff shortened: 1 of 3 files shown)


alejandrohdezma (approved):

Oliver-Taylor:
## Production validation result: no measurable win — closing

We validated this change on a metric-heavy production service (audience-matching-internal, ~600 req/s/pod, ~10–20 metric updates per request) by building the branch locally and deploying it twice, profiling each iteration with JFR under real traffic.

**Iteration 1 (as originally raised):** the cache worked as designed — the old resolution path (label rendering, array build, upstream `Labels` allocate/sort/hash) went from 3.2% of execution samples to zero — but the resolver itself cost **6.7%**, more than it removed. Frames pointed at `computeIfAbsent`, which we diagnosed as bin-locking for collision-chained keys.

**Iteration 2 (get-then-compute, `0e88041`):** deployed and verified the lock-free `get` path was active — and the cost was **unchanged at 6.7%**. The real cost is the typed-key hashing itself: `Statics.anyHash` walking the product and boxed `equals` per element on every lookup. Scala tuple/case-class hashing per update costs roughly what the string-rendering path it replaced did.

Conclusion: the datapoint lookup was never expensive enough to beat the cost of any per-call keying scheme. The remaining per-update overhead in the profile is the `F[Unit]` effect wrapping and error-handling machinery around each increment — which points at a different design entirely (windowed local accumulation with periodic flush, invisible to 15–60s scrape intervals) rather than caching resolution.

Closing rather than merging; the branch and these findings remain for reference if the accumulation approach is picked up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #151, opened 2026-08-13
Add withStaleSeriesEviction to JavaMetricRegistry.Builder

gruffyn:
Adds an opt-in `Builder.withStaleSeriesEviction(ttl)` to the v6 Java backend: the registry tracks the last write to every (metric, label values) pair, and a background sweep (every `ttl / 4`) removes label sets that have not been written to within `ttl` via `StatefulMetric.remove`, so idle series stop being exposed. A later write recreates the series, which restarts from zero.

Motivation: metrics labelled by unbounded or churning values otherwise accumulate dead series for the lifetime of the process, growing the exposition (and scrape cost) without bound. Eviction bounds it to the recently-active label sets.

Notes:
- Only labelled stateful metrics participate; unlabelled metrics and callback-backed metrics are untouched.
- A write racing a sweep keeps its tracking entry (conditional remove), so it is re-evaluated next sweep; at worst a series written in the instant of its eviction restarts from zero.
- No public API changes outside the new `Builder` method; the core `MetricRegistry` abstraction is untouched.

Tested with a suite case covering idle-series removal, survival of actively-written series, and recreation after eviction.

### DIFF
--- a/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
+++ b/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
@@ -29,6 +29,7 @@ import cats.syntax.all._
 import io.prometheus.metrics.core.datapoints.CounterDataPoint
 import io.prometheus.metrics.core.datapoints.DistributionDataPoint
 import io.prometheus.metrics.core.datapoints.GaugeDataPoint
+import io.prometheus.metrics.core.metrics.StatefulMetric
 import io.prometheus.metrics.core.metrics.{Counter => PCounter}
 import io.prometheus.metrics.core.metrics.{Gauge => PGauge}
 import io.prometheus.metrics.core.metrics.{Histogram => PHistogram}
@@ -38,6 +39,7 @@ import io.prometheus.metrics.instrumentation.jvm.JvmMetrics
 import io.prometheus.metrics.model.registry.PrometheusRegistry
 import io.prometheus.metrics.model.snapshots.Labels
 import prometheus4cats._
+import prometheus4cats.javaclient.internal.StaleSeriesEvictor
 import prometheus4cats.javaclient.internal.Utils
 import prometheus4cats.javaclient.models.MetricType
 import prometheus4cats.util.DoubleMetricRegistry
@@ -55,8 +57,22 @@ class JavaMetricRegistry[F[_]: Async] private (
     private val registry: PrometheusRegistry,
     private val ref: Ref[F, State[F]],
     private val sem: Semaphore[F],
-    private val logger: Throwable => String => F[Unit]
+    private val logger: Throwable => String => F[Unit],
+    private val evictor: Option[StaleSeriesEvictor]
 ) extends DoubleMetricRegistry[F] {
+
+  private def touched[D](
+      labelNames: IndexedSeq[Label.Name],
+      metric: StatefulMetric[_, _]
+  )(getDataPoint: Array[String] => D): Array[String] => D =
+    evictor match {
+      case Some(e) if labelNames.nonEmpty =>
+        lbls => {
+          e.touch(metric, lbls)
+          getDataPoint(lbls)
+        }
+      case _ => getDataPoint
+    }
 
   type Underlying = PrometheusRegistry
 
@@ -185,7 +201,7 @@ class JavaMetricRegistry[F[_]: Async] private (
             allLabelNames = allLabelNames,
             dynamicLabels = f(labels),
             commonLabelValues = commonLabelValuesArray,
-            getDataPoint = (lbls: Array[String]) => counter.labelValues(lbls: _*),
+            getDataPoint = touched(labelNames, counter)((lbls: Array[String]) => counter.labelValues(lbls: _*)),
             modify = (dp: CounterDataPoint) =>
               exemplar.fold(dp.inc(d))(e => dp.incWithExemplar(d, transformExemplarLabels(e))),
             logger = logger
@@ -227,7 +243,7 @@ class JavaMetricRegistry[F[_]: Async] private (
           allLabelNames = allLabelNames,
           dynamicLabels = f(labels),
           commonLabelValues = commonLabelValuesArray,
-          getDataPoint = (lbls: Array[String]) => gauge.labelValues(lbls: _*),
+          getDataPoint = touched(labelNames, gauge)((lbls: Array[String]) => gauge.labelValues(lbls: _*)),
           modify = g,
           logger = logger
         )
@@ -284,7 +300,7 @@ class JavaMetricRegistry[F[_]: Async] private (
             allLabelNames = allLabelNames,
             dynamicLabels = f(labels),
             commonLabelValues = commonLabelValuesArray,
-            getDataPoint = (lbls: Array[String]) => histogram.labelValues(lbls: _*),
+            getDataPoint = touched(labelNames, histogram)((lbls: Array[String]) => histogram.labelValues(lbls: _*)),
             modify = (dp: DistributionDataPoint) =>
               exemplar.fold(dp.observe(d))(e => dp.observeWithExemplar(d, transformExemplarLabels(e))),
             logger = logger
@@ -425,7 +441,7 @@ class JavaMetricRegistry[F[_]: Async] private (
           allLabelNames = allLabelNames,
           dynamicLabels = f(labels),
           commonLabelValues = commonLabelValuesArray,
-          getDataPoint = (lbls: Array[String]) => summary.labelValues(lbls: _*),
+          getDataPoint = touched(labelNames, summary)((lbls: Array[String]) => summary.labelValues(lbls: _*)),
           modify = (dp: DistributionDataPoint) => dp.observe(d),
           logger = logger
         )
@@ -539,15 +555,17 @@ object JavaMetricRegistry {
   sealed abstract class Builder[F[_]: Async](
       val promRegistry: PrometheusRegistry,
       val logger: Throwable => String => F[Unit],
-      val registerJvmMetrics: Boolean
+      val registerJvmMetrics: Boolean,
+      val staleSeriesTtl: Option[FiniteDuration]
   ) {
 
     private def copy(
         promRegistry: PrometheusRegistry = promRegistry,
         logger: Throwable => String => F[Unit] = logger,
-        registerJvmMetrics: Boolean = registerJvmMetrics
+        registerJvmMetrics: Boolean = registerJvmMetrics,
+        staleSeriesTtl: Option[FiniteDuration] = staleSeriesTtl
     ): Builder[F] =
-      new Builder(promRegistry, logger, registerJvmMetrics) {}
+      new Builder(promRegistry, logger, registerJvmMetrics, staleSeriesTtl) {}
 
     def withRegistry(promRegistry: PrometheusRegistry): Builder[F] = copy(promRegistry = promRegistry)
 
@@ -559,26 +577,45 @@ object JavaMetricRegistry {
       */
     def withJvmMetrics: Builder[F] = copy(registerJvmMetrics = true)
 
+    /** Evict label sets that have not been written to for `ttl`. A background sweep (every quarter of the TTL) removes
+      * idle series from the underlying collectors, so they stop being exposed; a later write recreates the series,
+      * which then starts again from zero. Intended for metrics labelled by unbounded or churning values, where the
+      * registry would otherwise accumulate dead series for the lifetime of the process. Only metrics with at least one
+      * label are affected; callback-backed metrics never are.
+      */
+    def withStaleSeriesEviction(ttl: FiniteDuration): Builder[F] = copy(staleSeriesTtl = Some(ttl))
+
     def build: Resource[F, JavaMetricRegistry[F]] =
       Resource.eval {
         if (registerJvmMetrics) Sync[F].delay(JvmMetrics.builder().register(promRegistry))
         else Applicative[F].unit
       }.flatMap { _ =>
+        val evictor = staleSeriesTtl.map(new StaleSeriesEvictor(_))
+
         val acquire = for {
           ref <- Ref.of[F, State[F]](Map.empty)
           sem <- Semaphore[F](1L)
-          reg  = new JavaMetricRegistry[F](promRegistry, ref, sem, logger)
+          reg  = new JavaMetricRegistry[F](promRegistry, ref, sem, logger, evictor)
         } yield reg
 
-        Resource.make(acquire) { reg =>
-          reg.ref.get.flatMap { metrics =>
-            if (metrics.nonEmpty)
-              metrics.values.toList.traverse_ { case (_, (collector, _, _)) =>
-                Utils.unregister(collector, promRegistry, logger)
-              }
-            else Applicative[F].unit
+        Resource
+          .make(acquire) { reg =>
+            reg.ref.get.flatMap { metrics =>
+              if (metrics.nonEmpty)
+                metrics.values.toList.traverse_ { case (_, (collector, _, _)) =>
+                  Utils.unregister(collector, promRegistry, logger)
+                }
+              else Applicative[F].unit
+            }
           }
-        }
+          .flatTap { _ =>
+            evictor.fold(Resource.unit[F]) { e =>
+              val sweep = Sync[F]
+                .delay(e.sweep())
+                .handleErrorWith(logger(_)("Failed to evict stale Prometheus series"))
+              Async[F].background((Temporal[F].sleep(e.sweepInterval) >> sweep).foreverM[Unit]).void
+            }
+          }
       }
 
   }
@@ -589,7 +626,8 @@ object JavaMetricRegistry {
       new Builder[F](
         promRegistry = PrometheusRegistry.defaultRegistry,
         logger = _ => _ => Async[F].unit,
-        registerJvmMetrics = false
+        registerJvmMetrics = false,
+        staleSeriesTtl = None
       ) {}
 
   }
--- a/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/internal/StaleSeriesEvictor.scala
+++ b/modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/internal/StaleSeriesEvictor.scala
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2022-2026 Permutive Ltd. <https://permutive.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package prometheus4cats.javaclient.internal
+
+import java.util.concurrent.ConcurrentHashMap
+
+import scala.concurrent.duration._
+
+import io.prometheus.metrics.core.metrics.StatefulMetric
+
+/** Tracks the last write time of every (metric, label values) pair and removes pairs that have not been written to
+  * within `ttl` from the underlying collector, so idle series stop being exposed. A series removed by a sweep is
+  * recreated (from zero) by its next write. Writes racing a sweep keep their map entry and are re-evaluated on the next
+  * sweep; at worst a series that was written to in the instant it was being evicted restarts from zero.
+  */
+final private[javaclient] class StaleSeriesEvictor(ttl: FiniteDuration) {
+
+  import StaleSeriesEvictor.Key
+
+  private[this] val lastTouched = new ConcurrentHashMap[Key, java.lang.Long]()
+
+  val sweepInterval: FiniteDuration = ttl / 4
+
+  def touch(metric: StatefulMetric[_, _], labelValues: Array[String]): Unit =
+    lastTouched.put(new Key(metric, labelValues), java.lang.Long.valueOf(System.nanoTime())): Unit
+
+  def sweep(): Unit = {
+    val cutoff = System.nanoTime() - ttl.toNanos
+    lastTouched.forEach { (key, lastWrite) =>
+      if (lastWrite.longValue() < cutoff && lastTouched.remove(key, lastWrite))
+        key.metric.remove(key.labelValues: _*)
+    }
+  }
+
+}
+
+private[javaclient] object StaleSeriesEvictor {
+
+  @SuppressWarnings(Array("scalafix:Disable.equals", "scalafix:Disable.hashCode"))
+  final private class Key(val metric: StatefulMetric[_, _], val labelValues: Array[String]) {
+
+    override val hashCode: Int =
+      System.identityHashCode(metric) * 31 + java.util.Arrays.hashCode(labelValues.asInstanceOf[Array[AnyRef]])
+
+    override def equals(other: Any): Boolean = other match {
+      case that: Key =>
+        (that.metric eq metric) && java.util.Arrays.equals(
+          that.labelValues.asInstanceOf[Array[AnyRef]],
+          labelValues.asInstanceOf[Array[AnyRef]]
+        )
+      case _ => false
+    }
+
+  }
+
+}
--- a/modules/prometheus4cats-java/src/test/scala/prometheus4cats/javaclient/JavaMetricRegistrySuite.scala
+++ b/modules/prometheus4cats-java/src/test/scala/prometheus4cats/javaclient/JavaMetricRegistrySuite.scala
@@ -16,6 +16,7 @@
 
 package prometheus4cats.javaclient
 
+import scala.concurrent.duration._
 import scala.jdk.CollectionConverters._
 
 import cats.effect.IO
@@ -155,5 +156,46 @@ class JavaMetricRegistrySuite extends CatsEffectSuite with DslSuite {
           )
       }
       .sortBy(_.name)
+
+  // ─── suite-local tests ────────────────────────────────────────────────────────────────────────────
+
+  test("withStaleSeriesEviction removes idle label sets and recreates them on the next write") {
+    val promRegistry = new PrometheusRegistry()
+
+    def series(labelValue: String): Option[Double] =
+      promRegistry
+        .scrape()
+        .asScala
+        .toList
+        .collectFirst { case s: CounterSnapshot =>
+          s.getDataPoints.asScala.collectFirst {
+            case dp if promLabelsToMap(dp.getLabels).get("status").contains(labelValue) => dp.getValue
+          }
+        }
+        .flatten
+
+    JavaMetricRegistry
+      .Builder[IO]()
+      .withRegistry(promRegistry)
+      .withStaleSeriesEviction(500.millis)
+      .build
+      .map(MetricFactory.builder.build[IO](_))
+      .flatMap(_.counter("eviction_test_total").help("eviction test").label[String]("status").build)
+      .use { counter =>
+        for {
+          _      <- counter.inc("idle") >> counter.inc("active")
+          before <- IO.delay((series("idle"), series("active")))
+          _      <- (IO.sleep(100.millis) >> counter.inc("active")).replicateA_(10)
+          after  <- IO.delay((series("idle"), series("active")))
+          _      <- counter.inc("idle")
+          revived = series("idle")
+        } yield {
+          assertEquals(before, (Some(1.0), Some(1.0)))
+          assertEquals(after._1, None)
+          assert(after._2.exists(_ >= 10.0))
+          assertEquals(revived, Some(1.0))
+        }
+      }
+  }
 
 }


modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -559,26 +577,45 @@ object JavaMetricRegistry {
...
+                metrics.values.toList.traverse_ { case (_, (collector, _, _)) =>
+                  Utils.unregister(collector, promRegistry, logger)
+                }
+              else Applicative[F].unit
+            }
+          }
+          .flatTap { _ =>
+            evictor.fold(Resource.unit[F]) { e =>
+              val sweep = Sync[F]
+                .delay(e.sweep())
```
kidonm: **issue** (blocking): The sweep runs on the compute pool and can starve a worker thread.

`Sync[F].delay(e.sweep())` is one uninterruptible block doing a full-map `forEach` over every tracked series, with a `remove` on the evictor's map *and* a `metric.remove(...)` against the collector's own `ConcurrentHashMap` for each expired entry. At the scale this feature exists for — a long TTL over churning labels can easily hold hundreds of thousands of entries — that's ~15-25ms of straight-line CPU holding a compute worker, and it happens with a 1-hour TTL just as much as a short one. The complexity class isn't the problem; *where it runs* is.

```scala
val sweep = Sync[F]
  .blocking(e.sweep())          // or interruptible, so cancellation doesn't wait on a full sweep
  .handleErrorWith(logger(_)("Failed to evict stale Prometheus series"))
```

(Moot if the scrape-driven approach in the `sweepInterval` comment wins, since eviction then runs on the exporter's own thread.)

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/internal/StaleSeriesEvictor.scala
```diff
@@ -0,0 +1,70 @@
...
+  * recreated (from zero) by its next write. Writes racing a sweep keep their map entry and are re-evaluated on the next
+  * sweep; at worst a series that was written to in the instant it was being evicted restarts from zero.
+  */
+final private[javaclient] class StaleSeriesEvictor(ttl: FiniteDuration) {
+
+  import StaleSeriesEvictor.Key
+
+  private[this] val lastTouched = new ConcurrentHashMap[Key, java.lang.Long]()
+
+  val sweepInterval: FiniteDuration = ttl / 4
```
kidonm: **suggestion** (blocking): Consider driving eviction from the scrape rather than a background timer.

The current shape needs a background fiber *and* a `sweepInterval`, and `ttl / 4` is an unexplained magic number. Since the thing we actually care about is "don't *expose* stale series", the scrape is the natural trigger — and the upstream API supports it. `Collector` is a `@FunctionalInterface`, `Metric.Builder` has `build()` separate from `register(registry)`, and `StatefulMetric` already has the right primitive (`StatefulMetric.java:141`):

```java
public void removeIf(Function<List<String>, Boolean> f) {
  data.entrySet().removeIf(entry -> f.apply(Collections.unmodifiableList(entry.getKey())));
}
```

So you can register a decorator instead of the metric itself:

```scala
final private[javaclient] class EvictingCollector(
    underlying: StatefulMetric[_, _] with MetricWithFixedMetadata,
    ttl: FiniteDuration
) extends Collector {

  private[this] val lastWrite = new ConcurrentHashMap[java.util.List[String], java.lang.Long]()

  def touch(labelValues: java.util.List[String], now: Long): Unit = ...

  def collect(): MetricSnapshot = {
    val snapshot = underlying.collect()   // expose one final sample, then evict
    val cutoff   = System.nanoTime() - ttl.toNanos
    underlying.removeIf { labels =>
      val t     = lastWrite.get(labels)
      val stale = (t ne null) && t.longValue() < cutoff
      if (stale) lastWrite.remove(labels, t): Unit
      stale
    }
    snapshot
  }

  override def getPrometheusName: String            = underlying.getPrometheusName
  override def getMetricType: MetricType            = underlying.getMetricType
  override def getMetadata: MetricMetadata          = underlying.getMetadata
  override def getLabelNames: java.util.Set[String] = underlying.getLabelNames
}
```

`touch` stays exactly where it is on the write path — only the sweep moves. What this removes: the background fiber, `sweepInterval`, `ttl / 4`, the compute-pool starvation issue, the custom `Key` class with its `identityHashCode` and two `scalafix:Disable` suppressions (a per-metric map keys on the same `List[String]` the collector itself uses), and the ordered-map concern — eviction now piggybacks on a scrape that is already O(series), on the exporter's own thread. Semantics also become exact: "not exposed if idle for `ttl`", evaluated at the moment of exposure, rather than "gone within `ttl + ttl/4`".

Three things worth weighing rather than waving through, in fairness to the current design:

1. **The delegation is load-bearing.** `PrometheusRegistry.register` only validates duplicate names, type conflicts, and help/unit consistency `if (prometheusName != null && metricType != null)`. A wrapper that misses any of those four overrides silently disables duplicate-metric detection for that metric.
2. **Real regression: an unscraped registry never evicts.** The background sweeper bounds memory whether or not anyone scrapes; scrape-driven eviction doesn't. Usually fine — an unscraped registry is a misconfiguration — but it's a deliberate trade.
3. **Plumbing cost.** The registration sites at lines 184, 232, 288, 430 use `builder.register(registry)` and would become `build()` + `registry.register(wrapper)`; `State[F]` and `Utils.unregister` are typed to `MetricWithFixedMetadata` and would widen to `Collector`. Contained, but not a one-liner.

Happy to be told the unscraped-registry case matters enough to keep the timer.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/internal/StaleSeriesEvictor.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  def touch(metric: StatefulMetric[_, _], labelValues: Array[String]): Unit =
+    lastTouched.put(new Key(metric, labelValues), java.lang.Long.valueOf(System.nanoTime())): Unit
+
+  def sweep(): Unit = {
+    val cutoff = System.nanoTime() - ttl.toNanos
+    lastTouched.forEach { (key, lastWrite) =>
+      if (lastWrite.longValue() < cutoff && lastTouched.remove(key, lastWrite))
+        key.metric.remove(key.labelValues: _*)
```
kidonm: **issue** (non-blocking): A write racing a sweep can have its increment silently discarded, leaving the map and the collector out of sync.

The CAS protects the *map entry*, but `metric.remove(...)` then runs unconditionally:

```
sweeper: CAS remove(key, oldTs) succeeds       → map entry gone
writer:  touch(...) → put(key, newTs)          → entry back, looks fresh
writer:  labelValues(...) → computeIfAbsent    → data point present
writer:  dp.inc(1)                             → value now 6
sweeper: metric.remove(labelValues: _*)        → data point deleted, the 6 is gone
```

The map now records the series as fresh at `newTs` while the collector has no data point for it — so a series that was *just written to* stays absent from exposition until its next write, which for a low-frequency series can be a long time. The scaladoc at lines 27-28 says racing writes "keep their map entry and are re-evaluated on the next sweep", which holds for the entry but not for the data point.

I don't think this is fully fixable with `remove(String...)` — the staleness decision can't be made atomic with the collector mutation. Two options: switch to `removeIf`, which at least evaluates the predicate inside the collector's own `entrySet().removeIf(...)` and narrows the window considerably; or keep the race and correct the comment to say what actually happens, e.g. "a write landing in the instant a sweep removes its data point may be lost, and the series will not be re-exposed until its following write."

Narrow window and low impact for a stale-series feature — flagging it mostly because the current comment promises more than the code delivers.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -559,26 +577,45 @@ object JavaMetricRegistry {
...
+                .handleErrorWith(logger(_)("Failed to evict stale Prometheus series"))
+              Async[F].background((Temporal[F].sleep(e.sweepInterval) >> sweep).foreverM[Unit]).void
```
kidonm: **issue** (non-blocking): If the sweep fiber dies, eviction is silently disabled for the rest of the process.

`logger` is user-supplied (`Throwable => String => F[Unit]`), so if it raises, `handleErrorWith` yields a failed effect, `foreverM` propagates it, and the fiber terminates. `.void` discards the `F[Outcome[...]]`, so nothing ever joins it — no log line, no crash, eviction just stops forever and memory grows as if the feature were off.

Cheapest fix is to make the loop unkillable by guarding the handler itself:

```scala
val sweep = Sync[F]
  .blocking(e.sweep())
  .handleErrorWith(e => logger(e)("Failed to evict stale Prometheus series").handleError(_ => ()))
```

Alternatively keep the outcome and log an errored termination. Either is fine; silently losing the sweeper is the part worth avoiding.

modules/prometheus4cats-java/src/test/scala/prometheus4cats/javaclient/JavaMetricRegistrySuite.scala
```diff
@@ -156,4 +157,45 @@ class JavaMetricRegistrySuite extends CatsEffectSuite with DslSuite {
...
+          _      <- counter.inc("idle") >> counter.inc("active")
+          before <- IO.delay((series("idle"), series("active")))
+          _      <- (IO.sleep(100.millis) >> counter.inc("active")).replicateA_(10)
+          after  <- IO.delay((series("idle"), series("active")))
+          _      <- counter.inc("idle")
+          revived = series("idle")
+        } yield {
+          assertEquals(before, (Some(1.0), Some(1.0)))
+          assertEquals(after._1, None)
+          assert(after._2.exists(_ >= 10.0))
```
kidonm: **issue** (non-blocking, test): This assertion will flake on a loaded CI box.

With a 500ms TTL, `active` is written every ~100ms and ends at 11, so `>= 10.0` leaves exactly one increment of slack. A single scheduling gap or GC pause over 500ms evicts `active`, recreates it from zero, and the count comes in under 10. The 1000ms of real sleeping also makes this one of the slower tests in the suite.

If `Clock[F].monotonic` lands per the comment on `StaleSeriesEvictor.scala:39`, `TestControl` removes the timing dependency entirely and the test becomes both deterministic and instant. Failing that, widening the TTL-to-write-interval ratio (say a 2s TTL with 100ms writes) buys a lot more headroom than the current 5×.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -185,7 +201,7 @@ class JavaMetricRegistry[F[_]: Async] private (
             allLabelNames = allLabelNames,
             dynamicLabels = f(labels),
             commonLabelValues = commonLabelValuesArray,
-            getDataPoint = (lbls: Array[String]) => counter.labelValues(lbls: _*),
+            getDataPoint = touched(labelNames, counter)((lbls: Array[String]) => counter.labelValues(lbls: _*)),
```
kidonm: **suggestion** (non-blocking): Hoist `touched(...)` out of the per-write lambda.

`touched(labelNames, counter)(...)` is evaluated *inside* the `_inc` lambda, so on every metric write it allocates the inner closure, pattern-matches on `evictor`, and allocates a second wrapping closure. Everything it closes over — `labelNames`, `counter`, `evictor` — is fixed at registration time, so this can be computed once. Right now the `case _ => getDataPoint` branch, which is clearly there to make the feature free when disabled, doesn't achieve that: you still pay a match and an allocation per write.

```scala
.map { case (counter, exemplarRef) =>
  val getDataPoint = touched(labelNames, counter)((lbls: Array[String]) => counter.labelValues(lbls: _*))

  Counter.make[F, Double, A](
    ...
    _inc = { (d, labels, exemplar) =>
      Utils.modifyMetric[F, Counter.Name, CounterDataPoint](..., getDataPoint = getDataPoint, ...)
    }
  )
}
```

Same at lines 246 (gauge), 303 (histogram), and 444 (summary).

Related and less fixable: `java.lang.Long.valueOf(System.nanoTime())` boxes outside the `-128..127` cache, so that's a third allocation per write. Avoiding it needs a primitive-valued map or a mutable holder, so probably just worth a comment rather than a change.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -559,26 +577,45 @@ object JavaMetricRegistry {
       */
     def withJvmMetrics: Builder[F] = copy(registerJvmMetrics = true)
 
+    /** Evict label sets that have not been written to for `ttl`. A background sweep (every quarter of the TTL) removes
+      * idle series from the underlying collectors, so they stop being exposed; a later write recreates the series,
+      * which then starts again from zero. Intended for metrics labelled by unbounded or churning values, where the
+      * registry would otherwise accumulate dead series for the lifetime of the process. Only metrics with at least one
+      * label are affected; callback-backed metrics never are.
+      */
+    def withStaleSeriesEviction(ttl: FiniteDuration): Builder[F] = copy(staleSeriesTtl = Some(ttl))
```
kidonm: **suggestion** (non-blocking): Guard against a non-positive `ttl`.

`sweepInterval` is `ttl / 4`, which truncates — a sufficiently small `FiniteDuration` gives a zero-length interval and turns the `foreverM` into a core-burning spin loop. A negative TTL would evict every series on the first sweep. Cheap to close off at the boundary:

```scala
def withStaleSeriesEviction(ttl: FiniteDuration): Builder[F] = {
  require(ttl > Duration.Zero, s"stale series TTL must be positive, got $ttl")
  copy(staleSeriesTtl = Some(ttl))
}
```

kidonm (commented):
Code review with 12 comments (2 blocking, 10 non-blocking).

Nice piece of work — the finalizer ordering in particular is subtle and correct: because `background` is attached via `.flatTap` after `Resource.make`, the sweep fiber is cancelled *before* the collectors are unregistered, so a sweep can never run against a half-torn-down registry. Easy to get wrong, invisible when right.

The main discussion below is about whether eviction should be driven by a background timer at all, versus lazily at scrape time via a decorating `Collector` — see the comment on `sweepInterval`.

---

**chore** (non-blocking): `withStaleSeriesEviction` is new public API with no CHANGELOG entry.

`CHANGELOG.md` isn't in this diff so I can't comment inline. The `## [Unreleased]` → `### Added` section wants a line, and the convention is active — both `bb21d50` and `27b90b7` updated it as part of the feature commit. Something like:

```markdown
- Opt-in stale-series eviction via `JavaMetricRegistry.Builder#withStaleSeriesEviction(ttl)`, which
  removes label sets that have not been written to within `ttl` so they stop being exposed
```

The operational caveats — memory retained per live label set, and the eviction-trigger semantics — would also be worth a paragraph in `website/docs/implementations/java.md`. Weaker ask, since `withJvmMetrics` isn't documented there either.

modules/prometheus4cats-java/src/test/scala/prometheus4cats/javaclient/JavaMetricRegistrySuite.scala
```diff
@@ -156,4 +157,45 @@ class JavaMetricRegistrySuite extends CatsEffectSuite with DslSuite {
       }
       .sortBy(_.name)
 
+  // ─── suite-local tests ────────────────────────────────────────────────────────────────────────────
+
+  test("withStaleSeriesEviction removes idle label sets and recreates them on the next write") {
```
kidonm: **suggestion** (non-blocking, test): Only `Counter` is covered; three other wirings and the no-label guard are untested.

`touched` is applied at four independent call sites — counter (204), gauge (246), histogram (303), summary (444) — and only the counter path is exercised. Each is a separate hand-written line that could reference the wrong metric or the wrong `labelNames`.

The gap I'd care about most is the `labelNames.nonEmpty` guard. A metric with no *dynamic* labels has exactly one series that must never be evicted; if that guard regressed it would thrash evict→recreate→evict forever, and nothing in the suite would notice. Worth an explicit test that a no-label counter still holds its value after several sweep intervals — and one that a metric whose only labels are *common* labels is likewise left alone, since `allLabelNames = labelNames ++ commonLabelNames` makes that distinction easy to get wrong.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/internal/StaleSeriesEvictor.scala
```diff
@@ -0,0 +1,70 @@
...
+final private[javaclient] class StaleSeriesEvictor(ttl: FiniteDuration) {
+
+  import StaleSeriesEvictor.Key
+
+  private[this] val lastTouched = new ConcurrentHashMap[Key, java.lang.Long]()
+
+  val sweepInterval: FiniteDuration = ttl / 4
+
+  def touch(metric: StatefulMetric[_, _], labelValues: Array[String]): Unit =
+    lastTouched.put(new Key(metric, labelValues), java.lang.Long.valueOf(System.nanoTime())): Unit
```
kidonm: **question** (non-blocking): Why raw `System.nanoTime()` rather than `Clock[F].monotonic`?

Two angles. On effectfulness: the call *is* suspended, just at the call site rather than here — `getDataPoint` runs inside `Sync[F].delay` at `internal/Utils.scala:67`, and `sweep()` is wrapped in `Sync[F].delay` in the builder. So nothing unsuspended escapes, and the class is `private[javaclient]`. Fine as far as it goes.

On testability it costs more: with the clock baked in, the evictor can't be driven by `TestControl` or a mock clock, which is why the accompanying test has to use real `IO.sleep` and wall-clock timings (and is timing-fragile as a result — separate comment).

Was `Clock[F]` considered and rejected? I can see the argument: `modifyMetric` takes `getDataPoint: Array[String] => D` as a *pure* function so the whole write is one `delay`, and threading `F` through would either add a `flatMap` per metric write or force a signature change. If that's the reasoning it's worth a comment here, because the next reader will ask the same question.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/internal/StaleSeriesEvictor.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  def sweep(): Unit = {
```
kidonm: **thought** (non-blocking): An ordered map would make expiry lookup logarithmic — but probably taxes the wrong path.

Worth thinking through explicitly, because the asymptotics are genuinely better and the conclusion isn't obvious. Ordering by *timestamp* rather than by `Key` gives you a `headMap` boundary in `O(log N)` and iteration over only the expired prefix:

```scala
private[this] val lastTouched = new ConcurrentHashMap[Key, java.lang.Long]()
private[this] val byExpiry    = new ConcurrentSkipListMap[(Long, Key), Unit]()

def touch(metric: StatefulMetric[_, _], labelValues: Array[String]): Unit = {
  val key  = new Key(metric, labelValues)
  val now  = System.nanoTime()
  val prev = lastTouched.put(key, now)
  byExpiry.put((now, key), ())
  if (prev ne null) byExpiry.remove((prev.longValue(), key)): Unit
}
```

`O(N)` per sweep becomes `O(log N + expired)`. But look at the cost to `touch`: one `O(1)` CHM put becomes a put *plus* a skip-list insert *plus* a skip-list removal, three concurrent-structure mutations and an extra composite-key allocation, on the hottest path in the library. And since the two maps can't be updated atomically you'd still need the sweep to validate against the CHM, so today's CAS stays.

The ratio is what settles it: with a 1-hour TTL the sweep runs four times an hour; `touch` runs on every metric write. So this moves work off a cold path onto a hot one. Recording it mainly so the trade-off is on the record rather than as a change request — and it's moot under the scrape-driven design, where `removeIf` walks the collector's own map and no ordering is needed at all.

modules/prometheus4cats-java/src/test/scala/prometheus4cats/javaclient/JavaMetricRegistrySuite.scala
```diff
@@ -156,4 +157,45 @@ class JavaMetricRegistrySuite extends CatsEffectSuite with DslSuite {
...
+    val promRegistry = new PrometheusRegistry()
+
+    def series(labelValue: String): Option[Double] =
```
kidonm: **nitpick** (non-blocking, test): Reuse the suite's existing full-state helper instead of a narrow value getter.

`series(labelValue): Option[Double]` pulls a single value out via `collectFirst`, while this file already has `scrapeToFamilyStates` producing a full `List[FamilyState]` with labels, values, and exemplars. Asserting whole-registry state would catch things this getter structurally cannot — the map/collector desync from the `StaleSeriesEvictor.scala:45` comment, or the evict/recreate thrash from the coverage comment above — and `collectFirst { case s: CounterSnapshot => ... }` silently picks the first counter, which is fine with one metric registered and quietly wrong the moment a second appears.

Also minor: line 191's `revived = series("idle")` is an unsuspended side effect inside the for-comprehension, inconsistent with the `IO.delay(...)` wrapping on lines 187 and 189.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/internal/StaleSeriesEvictor.scala
gruffyn: Adopted wholesale — `EvictingCollector` now decorates each labelled stateful metric and evicts via `removeIf` in `collect()`; the background fiber, `sweepInterval`, `Key` class and both suppressions are gone. On your three caveats: all four metadata accessors delegate (with a test pinning duplicate-name detection through the wrapper); the unscraped-registry regression is accepted as a misconfiguration and called out in the scaladoc, changelog and docs; the plumbing landed as `build()` + central registration in `configureBuilderOrRetrieve`, with the registered collector stored in the state so release unregisters the same object.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
gruffyn: Moot now — eviction runs in `collect()` on the exporter's thread.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/internal/StaleSeriesEvictor.scala
gruffyn: Switched to `removeIf` as part of the scrape-driven rework, and the scaladoc now states the residual window honestly: a write landing between the staleness check and the removal may be lost, and the series is not re-exposed until its following write.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
gruffyn: Moot — no background fiber any more.

modules/prometheus4cats-java/src/test/scala/prometheus4cats/javaclient/JavaMetricRegistrySuite.scala
gruffyn: Replaced with deterministic unit tests against `EvictingCollector` via an injectable `now: () => Long`; the remaining integration tests only ever sleep past the TTL (lower-bounded by `IO.sleep`, so no upper-bound timing to flake on) and the racing-writer loop is gone.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
gruffyn: Done — `getDataPoint` is now a `val` computed once per registration at all four sites, and the disabled path is the untouched function with no per-write match or allocation. The boxed-Long (and `Arrays.asList` wrapper) per-write cost is noted in the class scaladoc.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
gruffyn: Added, with `Duration.Zero`. `sweepInterval` no longer exists, but a non-positive TTL would still evict every touched series on each scrape, so the guard stays useful.

modules/prometheus4cats-java/src/test/scala/prometheus4cats/javaclient/JavaMetricRegistrySuite.scala
gruffyn: Coverage added: an integration test drives counter, gauge, histogram and summary through evict-and-recreate, and two guard tests pin that a no-label counter and a common-labels-only counter survive scrapes past the TTL.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/internal/StaleSeriesEvictor.scala
gruffyn: Your reading of the reasoning is right — `collect()` runs on the exporter's thread outside any effect context, and threading `Clock[F]` through would force `getDataPoint` effectful. It's now written down in the `EvictingCollector` scaladoc, and the clock is injectable (`now: () => Long`, defaulting to `System.nanoTime`) which is what the unit tests use in place of `TestControl`.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/internal/StaleSeriesEvictor.scala
gruffyn: Moot under the scrape-driven design — `removeIf` walks the collector's own map, so there's no separate expiry structure to order. Agreed on the hot-path arithmetic regardless.

modules/prometheus4cats-java/src/test/scala/prometheus4cats/javaclient/JavaMetricRegistrySuite.scala
gruffyn: The integration tests now assert through `scrapeToFamilyStates`, and the unsuspended `series(...)` call is gone with the old test.

gruffyn:
CHANGELOG entry added under Unreleased/Added, plus a "Stale Series Eviction" section in website/docs/implementations/java.md covering the operational caveats (scrape-driven trigger, unscraped-registry behaviour, per-label-set memory).

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/internal/EvictingCollector.scala
```diff
@@ -0,0 +1,84 @@
...
+  private[this] val lastWrite = new ConcurrentHashMap[java.util.List[String], java.lang.Long]()
+
+  def touch(labelValues: Array[String]): Unit =
+    lastWrite.put(java.util.Arrays.asList(labelValues: _*), java.lang.Long.valueOf(now())): Unit
+
+  @SuppressWarnings(Array("scalafix:DisableSyntax.null"))
+  override def collect(): MetricSnapshot = {
+    val snapshot = underlying.collect()
+    val cutoff   = now() - ttlNanos
+    underlying.removeIf { labels =>
```
kidonm: **suggestion** (non-blocking): Scan `lastWrite` rather than `underlying.removeIf` — fewer hashes per scrape, and it folds in a `nanoTime` comparison fix.

Upstream's `removeIf` iterates every live data point and applies the predicate to each:

```java
public void removeIf(Function<List<String>, Boolean> f) {
    data.entrySet().removeIf(entry ->
      f.apply(Collections.unmodifiableList(entry.getKey())));
}
```

So each scrape does N `List[String]` hashes plus N CHM lookups to find the K stale series — and K is normally 0. Flipping which map is scanned makes the common path N long comparisons, allocating only for the K series actually being removed. `lastWrite` is the same size as the live series set, since label sets never seen by `touch` are never evicted by design.

Two things come along for free. `t.longValue() < cutoff` compares `System.nanoTime()` values absolutely, but `nanoTime`'s origin is arbitrary and it's only documented to be meaningful when compared by *difference* — theoretical at ~292 years of uptime, but the subtraction form costs nothing. And gating the `underlying.remove` on a *successful* conditional `lastWrite.remove` narrows the lost-write race documented at lines 47-48: a racing write now keeps its data point instead of having it removed regardless.

One way this could look:

```scala
override def collect(): MetricSnapshot = {
  val snapshot = underlying.collect()
  val cutoff   = now() - ttlNanos
  val it       = lastWrite.entrySet().iterator()
  while (it.hasNext) {
    val entry = it.next()
    val key   = entry.getKey
    if (entry.getValue.longValue() - cutoff < 0 && lastWrite.remove(key, entry.getValue))
      underlying.remove(key.toArray(new Array[String](key.size())): _*)
  }
  snapshot
}
```

The cost is losing `removeIf`'s single-call tidiness for an explicit iterator loop. Nothing wrong with the current version's results — the scrape is already O(live series) in upstream's own `collect()` and in text encoding, so this is a constant-factor thing rather than a fix.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -559,6 +598,18 @@ object JavaMetricRegistry {
...
 
+    /** Evict label sets that have not been written to for `ttl`. Eviction is driven by the scrape: each scrape exposes
+      * a stale series one final time and removes it from the underlying collector, so it is absent from subsequent
+      * scrapes; a later write recreates the series, which then starts again from zero. Intended for metrics labelled by
+      * unbounded or churning values, where the registry would otherwise accumulate dead series for the lifetime of the
+      * process. Only metrics with at least one label are affected; callback-backed metrics never are. Because eviction
+      * only happens at scrape time, a registry that is never scraped never evicts.
+      */
+    def withStaleSeriesEviction(ttl: FiniteDuration): Builder[F] = {
+      require(ttl > Duration.Zero, s"stale series TTL must be positive, got $ttl")
```
kidonm: **question** (non-blocking): Was fail-fast here deliberate, rather than validating the TTL in `build`?

Every other `Builder` method is a pure `copy`, and `build` is where effects happen — so this is the only method on the type that can throw, and an `IllegalArgumentException` raised at construction time escapes the `Resource`/`F` error channel that callers would otherwise handle uniformly.

Fail-fast on what is unambiguously programmer error is defensible, so this may well be intentional. If you'd rather keep the builder total:

```scala
def withStaleSeriesEviction(ttl: FiniteDuration): Builder[F] = copy(staleSeriesTtl = Some(ttl))

def build: Resource[F, JavaMetricRegistry[F]] =
  Resource.eval {
    staleSeriesTtl.traverse_ { ttl =>
      Sync[F].raiseUnless(ttl > Duration.Zero)(
        new IllegalArgumentException(s"stale series TTL must be positive, got $ttl")
      )
    }
  } >> existingBuild
```

kidonm (commented):
Code review with 2 comments (0 blocking, 2 non-blocking).

Nice piece of work — the snapshot-then-evict ordering, the touch-before-lookup ordering on the write path, and the note about why every metadata accessor has to delegate are all the non-obvious details, and they're all handled and documented. Verified that all four `labelValues` call sites go through `withTouch`, so there's no write path that mutates a series without refreshing its timestamp, and that the wrapper is unregistered under the same identity it was registered with.

Both comments are non-blocking: one constant-factor improvement to the eviction scan, one question about builder validation.

kidonm (approved):
LGTM % I think this would benefit from more eyes. @alejandrohdezma I think you'd be best positioned to give this a look please.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -55,9 +58,18 @@ class JavaMetricRegistry[F[_]: Async] private (
...
-    private val logger: Throwable => String => F[Unit]
+    private val logger: Throwable => String => F[Unit],
+    private val staleSeriesTtl: Option[FiniteDuration]
 ) extends DoubleMetricRegistry[F] {
 
+  private def withTouch[D](
+      evicting: Option[EvictingCollector]
+  )(getDataPoint: Array[String] => D): Array[String] => D =
+    evicting.fold(getDataPoint) { e => lbls =>
+      e.touch(lbls)
```
alejandrohdezma: **issue** (blocking): A rejected write leaks an entry in `lastWrite` that eviction can never reclaim.

`touch` runs before `getDataPoint`, so the label set is recorded in `EvictingCollector.lastWrite` even when `labelValues` throws and no data point is created. `StatefulMetric.labelValues` rejects a `null` label value from inside `computeIfAbsent`, and `Label.Value.apply` has no null check, so this is reachable from the public DSL with `.label[String]("user")(identity)` and a null input. `Utils.modifyMetric` logs and swallows the exception.

Because `collect()` evicts via `underlying.removeIf`, the predicate only visits keys present in the metric's own data map. A `lastWrite` key with no matching data point is never visited, so it lives for the process lifetime, one entry per distinct failing label set. That is the churning-label workload this feature exists for.

Reordering closes it, since a failed lookup then never touches:

```scala
evicting.fold(getDataPoint) { e => lbls =>
  val d = getDataPoint(lbls)
  e.touch(lbls)
  d
}
```

If the `lastWrite`-scan suggestion in the other thread is adopted, that also reclaims such orphans on the next scrape, but the reorder is worth doing regardless so they are not created in the first place.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -115,19 +133,31 @@ class JavaMetricRegistry[F[_]: Async] private (
             case None =>
               for {
                 exemplarRef <- Ref.of[F, Option[Exemplar.Data]](None)
-                collector   <- Sync[F].delay(register())
-              } yield (metrics.updated(fullName, (metricId, (collector, exemplarRef, 1))), (collector, exemplarRef))
+                registered <- Sync[F].delay {
+                                val metric = build()
+                                val evicting = staleSeriesTtl
+                                  .filter(_ => dynamicLabels.nonEmpty)
```
alejandrohdezma: **issue** (blocking): Labelled set-once gauges are evicted too, and nothing warns about it.

The wrap decision only checks `dynamicLabels.nonEmpty`; `metricType` is in scope here but never consulted. So a registry that enables eviction for a tenant-labelled counter also evicts `build_info{version="1.2.3"} = 1` set once at startup, or `queue_depth{shard}` set only on change. After the TTL the next scrape exposes the gauge one final time and then removes it, and any `absent(...)` or `== 1` alert misfires until the next `set`.

The Builder scaladoc, `website/docs/implementations/java.md` and the CHANGELOG only warn about counters restarting from zero, which is the benign case. A gauge that silently disappears is the surprising one.

At minimum the docs should say that labelled gauges are evicted as well. Better would be to exclude `MetricType.GAUGE` from wrapping, or accept a predicate on `(name, metricType)` so callers can scope eviction to the metrics that actually churn. Either is a small change at this site.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
```diff
@@ -559,6 +598,18 @@ object JavaMetricRegistry {
       */
     def withJvmMetrics: Builder[F] = copy(registerJvmMetrics = true)
 
+    /** Evict label sets that have not been written to for `ttl`. Eviction is driven by the scrape: each scrape exposes
+      * a stale series one final time and removes it from the underlying collector, so it is absent from subsequent
+      * scrapes; a later write recreates the series, which then starts again from zero. Intended for metrics labelled by
+      * unbounded or churning values, where the registry would otherwise accumulate dead series for the lifetime of the
+      * process. Only metrics with at least one label are affected; callback-backed metrics never are. Because eviction
```
alejandrohdezma: **nitpick** (non-blocking): "at least one label" should be "at least one dynamic label".

The code checks `dynamicLabels.nonEmpty` and `java.md` says the same, so a metric whose only labels are common labels is not affected (the test `leaves metrics without dynamic labels untouched` pins this). A user with `withCommonLabels(...)` and an unlabelled counter would read this scaladoc and expect eviction that never happens.

While here: labelled `Info` metrics go through `createAndRegisterInfo`, which still calls `.register(registry)` directly and is never wrapped. Worth a mention alongside callback-backed metrics, here and in `java.md`.

alejandrohdezma (commented):
Code review with 3 comments (2 blocking, 1 non-blocking).

The scrape-driven rework reads well and the earlier threads look properly closed. I have not repeated what is already open on the `lastWrite` scan and the `require` placement; both of those I would accept as proposed. The comments below are new.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/internal/EvictingCollector.scala
gruffyn: Adopted. `collect()` now scans `lastWrite`: comparison by difference, and the `underlying.remove` is gated on a successful conditional `lastWrite.remove`, so a racing write keeps its data point — the scaladoc now says what is lost is the eviction, re-evaluated on the next scrape.

Used `forEach` rather than an explicit iterator because `.scalafix.conf` sets `noWhileLoops`/`noVars`; CHM `forEach` is weakly consistent and permits removal mid-traversal, so it is equivalent. The `ne null` check and its `DisableSyntax.null` suppression go with the flip.

The scan also reclaims the orphaned entries from Alejandro's thread; new unit test covers that.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
gruffyn: Deliberate, but not worth defending — moved. `withStaleSeriesEviction` is a pure `copy` again and `build` raises via `raiseUnless`, with a test for the relocated failure mode.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
gruffyn: Confirmed and fixed by the reorder. Reproduced against 1.6.1: `labelValues` throws `IllegalArgumentException: null label value for metric t and label user` before any data point exists, and does so on every attempt for that label set — and `Label.Value.apply` has no null check, so it is reachable exactly as you describe.

`withTouch` now looks up first and touches after. Both stay inside the same `Sync[F].delay`, and a `collect()` interleaving between data-point creation and the touch finds no tracking entry, so it will not evict. Marek's `lastWrite` scan is in as well, so orphans created before this are reclaimed too.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
gruffyn: Both fixed. The scaladoc now says "at least one *dynamic* label", matching the code and `java.md`, and `Info` is listed alongside callback-backed and common-labels-only metrics in both places.

Info's exclusion is structural rather than an oversight: upstream `Info` extends `MetricWithFixedMetadata`, so it has no `removeIf` and `EvictingCollector`'s `StatefulMetric` bound will not take it. It does have `remove(String...)`, so a separate wrapper could cover it later if there is demand.

modules/prometheus4cats-java/src/main/scala/prometheus4cats/javaclient/JavaMetricRegistry.scala
gruffyn: Documented rather than excluded. The Builder scaladoc, `java.md` and the CHANGELOG now state that every labelled stateful metric is evicted, gauges included, and spell out the set-once case: the gauge disappears until its next `set`, so `absent(...)` or `== 1` alerts over it can misfire — keep such gauges unlabelled, or do not enable eviction on the registry holding them.

Not excluding `GAUGE`, because per-tenant/per-shard gauges are a legitimate eviction target and silently skipping them surprises in the other direction. The `(name, metricType)` predicate is the principled knob for having both on one registry — happy to add it, but as a follow-up rather than growing the public API (and promoting `MetricType`) in this PR. Say the word if you would rather it landed here.

alejandrohdezma (approved):

